PR #1, opened 2017-04-24
Update the make file to be git-centric.

ericsnowcurrently:
As part of this we factor out a separate make file just for packaging, so that the main make file isn't so cluttered.

Note that the hard-coded version info (in landscape/__init__.py and debian/changelog) are reset back to the clean values they were supposed to be at.


Makefile
```diff
@@ -92,7 +79,7 @@ clean:
 	-rm -rf _trial_temp
 	-rm -rf docs/api
 	-rm -rf man/\*.1
-	-rm -rf sdist
+	$(MAKE) -f Makefile.packaging clean
```
simpoir: I think you may have to move this up (and add a comment to that effect) as that other makefile will call python imports in its header and generate a bunch of pyc files after the find runs.

This would brake sourcepackage.

Makefile.packaging
```diff
@@ -0,0 +1,67 @@
...
+.PHONY: sourcepackage
+sourcepackage: clean origtarball prepchangelog updateversion
+	# need to remove sdist here because it doesn't exist in the
+	# orig tarball
+	rm -rf sdist
+	debuild -S $(DEBUILD_OPTS)
+
+.PHONY: releasetarball
+releasetarball:
+	$(MAKE) sdist TARBALL_VERSION=$(UPSTREAM_VERSION)
```
simpoir: you can't call make sdist, the target is in this file not in the Makefile.

Makefile
ericsnowcurrently: ack

Makefile.packaging
ericsnowcurrently: ack

Makefile
```diff
@@ -149,28 +112,27 @@ run:
...
+.PHONY: etags
 etags:
 	-etags --languages=python -R .
 
+.PHONY: package
+package:  ## Generate the debian packages (use with DEBUILD_OPTS="-us -uc").
+	$(MAKE) -f Makefile.packaging $@
+
+.PHONY: sourcepackage
+sourcepackage:
```
ericsnowcurrently: good point; I got a few things mixed up here

ericsnowcurrently:
@CanonicalLtd/landscape

PR #6, opened 2017-04-26
Un-skip and fix the temporarily skipped tests.

ericsnowcurrently:
A few commits back we skipped a few tests to get CI passing.  This patch fixes those tests so that they pass now.


ericsnowcurrently:
@CanonicalLtd/landscape

PR #7, opened 2017-04-26
Constrain the landscape-common version.

ericsnowcurrently:
(fixed https://bugs.launchpad.net/landscape-client/+bug/1686532)

Without this change, upgrading from earlier versions won't work.


ericsnowcurrently:
@CanonicalLtd/landscape

PR #8, opened 2017-04-27
Fix packaging after coverage and README change.

simpoir:


debian/landscape-client.docs
```diff
@@ -1,2 +1,2 @@
-README
+README.md
```
ericsnowcurrently: This should not be necessary.  README exists (README.md is only a symlink to it for the sake of GH).

PR #11, opened 2017-07-18
Autoremove reporting

simpoir:
This changes adds client bits to report autoremovable packages back to landscape-server so it can act on them. lp:1208393

Testing: 

sudo apt-get install sl && sudo apt-mark auto sl && apt-get autoremove --dry-run
sudo ./scripts/landscape-client -c root-client.conf
accept in landscape, wait forever for package data to sync or not (server doesn't know about that field)
/tmp/landscape-root/package-reporter.log should queue autoremovable once (those are diffs)
grep autoremovable /tmp/landscape-root/broker.log # shows package report being sent even if currently ignored by servers

### DIFF
--- a/landscape/package/tests/test_reporter.py
+++ b/landscape/package/tests/test_reporter.py
@@ -90,6 +90,8 @@ class PackageReporterAptTest(LandscapeTest):
         self.set_pkg1_installed()
         self.facade.reload_channels()
         name1 = sorted(self.facade.get_packages_by_name("name1"))[0]
+        # Since no other package depends on this, all that's needed
+        # to have it autoremovable is to mark it as installed+auto.
         name1.package.mark_auto(True)
 
     def _make_fake_apt_update(self, out="output", err="error", code=0):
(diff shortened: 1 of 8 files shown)


landscape/package/tests/test_reporter.py
```diff
@@ -926,6 +933,51 @@ def got_result(result):
...
+        self.assertMessages(message_store.get_pending_messages(), expected)
+        self.assertEqual([1], self.store.get_autoremovable())
+
+    def test_detect_packages_changes_with_not_autoremovable(self):
+        message_store = self.broker_service.message_store
+        message_store.set_accepted_types(["packages"])
+
+        self.store.set_hash_ids({HASH1: 1, HASH2: 2, HASH3: 3})
+        self.store.add_available([1, 2, 3])
+        self.store.add_autoremovable([1, 2])
```
ericsnowcurrently: It may be worth adding a note here pointing out that add_installed() is never called, which is why not-autoremovable is what it is.

FWIW, I also find this approach to exercising not-autoremovable to be a little odd.  At what point would the store mark a package as auto-removable but not installed?

simpoir: Don't confuse the client store state with the apt package status. Unless the apt package is marked auto and not depended upon, it would still be flagged as not-autoremovable. It's true that in practice the installed state will come with autoremovable state in the store, but I don't care about exercising that path in this test.

I'll add a comment. on the logic

ericsnowcurrently: sounds good

PR #12, opened 2017-07-19
try fixing travis build with apt

simpoir:
try using apt to make build work on travis/trusty

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,10 @@ pipinstallpythonapt_deps:
 pipinstallpythonapt_src_trusty:
 	# travis-ci uses a now uses a backported version of apt.
 	# This matches the libapt-pkg-dev on those nodes so it compiles.
-	pip install https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
+	wget -O /tmp/python-apt.tar.xz https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
+	# py2.7 does not support .xz
+	xz -dk /tmp/python-apt.tar.xz
+	pip install /tmp/python-apt.tar
 
 .PHONY: pipinstallpythonapt_src_xenial
 pipinstallpythonapt_src_xenial:
--- a/Makefile
+++ b/Makefile
@@ -21,29 +21,6 @@ depends2:
 depends3:
 	sudo apt -y install python3-twisted python3-distutils-extra python3-mock python3-configobj python3-passlib
 
-.PHONY: pipinstallpythonapt
-pipinstallpythonapt: pipinstallpythonapt_deps
-	$(MAKE) pipinstallpythonapt_src_$(UBUNTU_RELEASE)
-
-.PHONY: pipinstallpythonapt_deps
-pipinstallpythonapt_deps:
-	pip install pyopenssl
-	pip install service_identity
-	sudo apt-get -y build-dep python-apt python3-apt
-
-.PHONY: pipinstallpythonapt_src_trusty
-pipinstallpythonapt_src_trusty:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -23,19 +23,22 @@ depends3:
...
+pipinstallpythonapt_src_trusty:
+	# travis-ci uses a now uses a backported version of apt.
+	# This matches the libapt-pkg-dev on those nodes so it compiles.
+	wget -O /tmp/python-apt.tar.xz https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
+	# py2.7 does not support .xz
+	xz -dk /tmp/python-apt.tar.xz
+	pip install /tmp/python-apt.tar
 
 .PHONY: pipinstallpythonapt_src_xenial
 pipinstallpythonapt_src_xenial:
```
ericsnowcurrently: Doesn't this target have the same py2.7 issue as the one you fixed in pipinstallpythonapt_src_trusty?  If so, it would make sense to fix this target too.

Makefile
```diff
@@ -23,19 +23,22 @@ depends3:
...
-.PHONY: pipinstallpythonapt_default
-pipinstallpythonapt_default:
-	# See: https://code.launchpad.net/ubuntu/+source/python-apt
-	bzr branch lp:ubuntu/$(UBUNTU_RELEASE)/python-apt /tmp/python-apt
-	pip install /tmp/python-apt
+.PHONY: pipinstallpythonapt_src_trusty
+pipinstallpythonapt_src_trusty:
+	# travis-ci uses a now uses a backported version of apt.
+	# This matches the libapt-pkg-dev on those nodes so it compiles.
+	wget -O /tmp/python-apt.tar.xz https://launchpad.net/ubuntu/+archive/primary/+files/python-apt_1.1.0~beta1build1.tar.xz
```
ericsnowcurrently: Doesn't Trusty use an older version?

Makefile
```diff
@@ -23,19 +23,22 @@ depends3:
...
 	sudo apt-get -y build-dep python-apt python3-apt
 
-.PHONY: pipinstallpythonapt_default
-pipinstallpythonapt_default:
-	# See: https://code.launchpad.net/ubuntu/+source/python-apt
-	bzr branch lp:ubuntu/$(UBUNTU_RELEASE)/python-apt /tmp/python-apt
-	pip install /tmp/python-apt
+.PHONY: pipinstallpythonapt_src_trusty
+pipinstallpythonapt_src_trusty:
+	# travis-ci uses a now uses a backported version of apt.
```
ericsnowcurrently: s/uses a now uses/now uses/

Makefile
```diff
@@ -23,19 +23,22 @@ depends3:
 
 .PHONY: pipinstallpythonapt
 pipinstallpythonapt: pipinstallpythonapt_deps
```
ericsnowcurrently: [optional]

It took me a minute to realize that the pipinstallpythonapt* targets are only used for Travis, where "pip" means the pip in a virtualenv targeting the right Python version.  While it can wait, consider pulling these targets out into a separate, Travis-specific make file.

It may also make sense to be explicit about the Python version in the various calls to pip:

    $(PYTHON) -m pip install ...

This would require that we set PYTHON=... when running make in .travis.yml, as we are in the script at the bottom of that file.

Makefile
```diff
@@ -23,19 +23,22 @@ depends3:
...
-	$(MAKE) pipinstallpythonapt_default || $(MAKE) pipinstallpythonapt_src_$(UBUNTU_RELEASE)
+	$(MAKE) pipinstallpythonapt_src_$(UBUNTU_RELEASE)
```
simpoir: good point. That'll make building slighly longer because of the failure, but it makes sense to at least try it

PR #14, opened 2017-09-15
Check if ubuntu-release-upgrader is running before apt-update.

dargad:
Fixes bug #1699179.

### DIFF
--- a/landscape/package/reporter.py
+++ b/landscape/package/reporter.py
@@ -230,8 +230,7 @@ class PackageReporter(PackageTaskHandler):
 
                 pid = os.path.basename(os.path.dirname(cmdline))
 
-                cmdline = [f for f in filter(lambda x: x != '',
-                                             read.split("\x00"))]
+                cmdline = [f for f in read.split("\x00") if f]
                 if len(cmdline) <= 1:
                     continue
 
@@ -242,15 +241,16 @@ class PackageReporter(PackageTaskHandler):
                                      re.VERBOSE | re.MULTILINE)
 
                 for pattern in pattern.finditer(read):
-                    uuid = pattern.groups()[0].split("\t")[1]
+                    uid = pattern.groups()[0].split("\t")[1]
             except IOError:
                 continue
 
             (executable, args) = (cmdline[0], cmdline[1:])
 
-            if (PYTHON_BIN in executable and
-                    any(RELEASE_UPGRADER_PATTERN in x for x in args) and
-                    uuid == UID_ROOT):
+            if (executable.startswith(PYTHON_BIN) and
+                    any(x.startswith(RELEASE_UPGRADER_PATTERN)
+                        for x in args) and
+                    uid == UID_ROOT):
                 logging.info("Found ubuntu-release-upgrader running (pid: %s)"
                              % (pid))
                 return True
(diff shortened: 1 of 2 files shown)


landscape/package/reporter.py
```diff
@@ -207,6 +211,51 @@ def _apt_update_timeout_expired(self, interval):
...
+        for cmdline in glob.glob("/proc/*/cmdline"):
+            base = os.path.dirname(cmdline)
+            try:
+                with open(cmdline) as fd:
+                    read = fd.read()
+
+                pid = os.path.basename(os.path.dirname(cmdline))
+
+                cmdline = [f for f in filter(lambda x: x != '',
+                                             read.split("\x00"))]
```
simpoir: I'd drop the lambda for list condition
cmdline = [f for f in read.split("\x00") if f]

landscape/package/reporter.py
```diff
@@ -207,6 +211,51 @@ def _apt_update_timeout_expired(self, interval):
...
+                                     re.VERBOSE | re.MULTILINE)
+
+                for pattern in pattern.finditer(read):
+                    uuid = pattern.groups()[0].split("\t")[1]
+            except IOError:
+                continue
+
+            (executable, args) = (cmdline[0], cmdline[1:])
+
+            if (PYTHON_BIN in executable and
```
simpoir: executable.startswith(PYTHON_BIN) would be better since you're specifying the path.

landscape/package/reporter.py
```diff
@@ -207,6 +211,51 @@ def _apt_update_timeout_expired(self, interval):
...
+                    any(RELEASE_UPGRADER_PATTERN in x for x in args) and
```
simpoir: same here. If we know the full path, I think startswith would be slightly faster to evaluate.

landscape/package/reporter.py
```diff
@@ -207,6 +211,51 @@ def _apt_update_timeout_expired(self, interval):
...
+                    continue
+
+                with open(os.path.join(base, "status")) as fd:
+                    read = fd.read()
+
+                pattern = re.compile('^Uid\:(.*)$',
+                                     re.VERBOSE | re.MULTILINE)
+
+                for pattern in pattern.finditer(read):
+                    uuid = pattern.groups()[0].split("\t")[1]
```
simpoir: I think uuid is a misnomer. uid would be better.

dargad:
I don't think the failing Python 3.6 lint checks have anything to do with changes introduced.
I'd appreciate shedding some light on that part. Thanks!

simpoir:
lint errors seem overly agressive on this build. The changes are fine, so I'll force-merge them.

landscape/package/reporter.py
```diff
@@ -28,6 +29,9 @@
 HASH_ID_REQUEST_TIMEOUT = 7200
 MAX_UNKNOWN_HASHES_PER_REQUEST = 500
 LOCK_RETRY_DELAYS = [0, 20, 40]
+PYTHON_BIN = "/usr/bin/python3"
```
panlinux: Will this be true even for older versions of ubuntu, like trusty? That the python interpreter will be python3?

dargad: I did the testing starting from Trusty 14.04 (the oldest version, without suffixes) and python3 is already installed there (as a dependency of ubuntu-minimal) and after running do-release-upgrade the release-upgrader that has been downloaded was using python3.
Since that's true for the oldest supported release I thought it's a safe assumption then.

PR #15, opened 2017-11-09
Isolate landscape.lib from the rest of the code.

ericsnowcurrently:
Currently modules under landscape.lib have some imports of modules outside of landscape.lib.  This patch fixes that by pulling relevant dependencies into landscape.lib.  Doing so matters for a number of reasons.  First, coupling between package trees increases maintenance burden and potentially increases the risk of defects.  Second, three independent code bases rely on landscape.lib: landscape-sysinfo, landscape-client, and landscape-server.  Coupling landscape.lib with the client code means that the full client codebase must be deployed with sysinfo and server (unnecessarily).  This wastes space and increases maintenance burden for the distro (since sysinfo will be installed by default).

Note that this PR is huge.  However, I've been careful to not make changes to the code (other than moving it and factoring out a couple of small pieces) and to maintain the existing test coverage.  So there shouldn't be much need to check every line.  Let me know if you have any concerns with this. :)

Here are the things I moved:

    landscape.compat -> landscape.lib.compat
    landscape.deployment (partial) -> landscape.lib.config
        ConfigSpecOptionParser
        BaseConfiguration (most of it)
        get_bindir
    landscape.log (partial) -> landscape.lib.format
        format_object
        format_delta
        format_percent
    landscape.monitor.computeruptime (partial) -> landscape.lib.sysstats
        get_uptime
        LoginInfo
        LoginInfoReader
        BootTimes
    landscape.reactor -> landscape.lib.reactor (with 2 exceptions)
        FakeReactor -> landscape.lib.testing.FakeReactor
        LandscapeReactor (mostly) -> EventHandlingReactor
    landscape.tests.helpers (partial) -> landscape.lib.testing
        HelperTestCase
        LandscapeTest (parts) -> CompatTestCase
        LandscapeTest (parts) -> ConfigTestCase
        LandscapeTest (parts) -> FSTestCase
        LandscapeTest (parts) -> TwistedTestCase
        LogKeeperHelper (and friends)
        EnvironSnapshot
        EnvironSaverHelper
        MockPopen
        StandardIOHelper
        mock_counter
        mock_time
        StubProcessFactory
        DummyProcess
        ProcessDataBuilder

I updated imports and tests accordingly.

Here are things that I added (factored out):

    landscape.lib.cli.add_cli_options()
    landscape.lib.config.add_cli_options()


landscape/lib/format.py
```diff
@@ -0,0 +1,31 @@
+import inspect
+
+
+def format_object(object):
```
simpoir: I was going to complain about shadowing builtin object, but I see this is verbatim of existing code… carry on.

PR #21, opened 2017-12-06
Fix the debian packaging.

ericsnowcurrently:
The packaging still has a few inconsistencies that produce warnings.  The Python-related bits were also not quite right.  Finally, I've switched to using variables for paths to make it easier to update later (and for better readability).


ericsnowcurrently:
Right.  I split them apart because I was planning on dropping the other one (#20).

PR #25, opened 2018-01-15
Reset package version to 18.01 pending bionic update.

simpoir:


debian/changelog
```diff
@@ -1,8 +1,8 @@
-landscape-client (17.04-0ubuntu1) UNRELEASED; urgency=medium
+landscape-client (18.01) UNRELEASED; urgency=medium
```
panlinux: Make it 18.01-0ubuntu1 please

PR #27, opened 2018-01-19
Best-effort decode old py27 message keys.

simpoir:
Testing instructions:

* apt-install landscape-client. (py27)
* configure in /etc/landscape/client.conf
* call /usr/bin/landscape-client and wait to be messages pending delivery
  * check /var/lib/landscape/client/messages/0 for a couple messages
* run ./script/landscape-broker -c /etc/landscape/client.conf \
    -d /var/lib/landscape/client
* Messages get delivered, broker queue gets emptied, no stack trace on client.
* Check on message-server for delivered messages.

### DIFF
--- a/landscape/client/broker/store.py
+++ b/landscape/client/broker/store.py
@@ -268,6 +268,17 @@ class MessageStore(object):
                 logging.exception(e)
                 self._add_flags(filename, BROKEN)
             else:
+                if u"type" not in message:
+                    # Special case to decode keys for messages which were
+                    # serialized by py27 prior to py3 upgrade, and having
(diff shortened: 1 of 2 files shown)


panlinux:
I built a package and tested a landscape-driven client upgrade (i.e., registered the py2 version with hosted, added a ppa for the py3 version, and let landscape itself drive the upgrade). It worked. I'll probably run some more specific tests next, but that's a good sign.

simpoir:
I've got also a couple of tests I'm running in hopes to to not find any message that gets badly reinterpreted.

PR #31, opened 2018-01-29
Implicitly trust file-local sources managed by landscape (LP: #1736576)

simpoir:
Testing instructions:

* install landscape-client on artful+ and register
* apply a simple package profile (depends sl)
* notice failure to update packages (Release file not found)
* apt-get build-depends ./ && dpkg-buildpackage -b --us -uc && dpkg -i ../*.deb
* check upgrade added trust option to :
  /etc/apt/sources.list.d/_landscape-internal-facad.list[.orig]
* re-check "apt-get update" works

### DIFF
--- a/debian/landscape-client.postinst
+++ b/debian/landscape-client.postinst
@@ -109,10 +109,9 @@ END
 
         # To work around bug #1735100 we rewrite file-local landscape sources
         # with the trusted flag, as they have no release file, thus are
-        # unsigned repositories.
+        # unsigned repositories. It exists while package profile is applying.
         LANDSCAPE_INTERNAL_SOURCES=/etc/apt/sources.list.d/_landscape-internal-facade.list
         if grep -q -e "^deb file:" $LANDSCAPE_INTERNAL_SOURCES; then
-            cp $LANDSCAPE_INTERNAL_SOURCES $LANDSCAPE_INTERNAL_SOURCES.orig
             sed -i 's/^deb file:/deb [ trusted=yes ] file:/' $LANDSCAPE_INTERNAL_SOURCES
         fi
     ;;
--- a/debian/landscape-client.postinst
+++ b/debian/landscape-client.postinst
@@ -106,6 +106,15 @@ END
         USER_UPDATE_FLAG_FILE="$DATA_PATH/user-update-flag"
         install --owner=landscape /dev/null $USER_UPDATE_FLAG_FILE
         echo "This file indicates that the Landscape client needs to send updated user information to the server." >> $USER_UPDATE_FLAG_FILE
+
+        # To work around bug #1735100 we rewrite file-local landscape sources
+        # with the trusted flag, as they have no release file, thus are
+        # unsigned repositories.
+        LANDSCAPE_INTERNAL_SOURCES=/etc/apt/sources.list.d/_landscape-internal-facade.list
+        if grep -q -e "^deb file:" $LANDSCAPE_INTERNAL_SOURCES; then
+            cp $LANDSCAPE_INTERNAL_SOURCES $LANDSCAPE_INTERNAL_SOURCES.orig
+            sed -i 's/^deb file:/deb [ trusted=yes ] file:/' $LANDSCAPE_INTERNAL_SOURCES
+        fi
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
--- a/landscape/lib/apt/package/facade.py
+++ b/landscape/lib/apt/package/facade.py
@@ -276,11 +276,11 @@ class AptFacade(object):
         @param url: The base URL of the repository.
         @param codename: The dist in the repository.
         @param components: The components to be included.
-        @param trusted: Whether validation shoule be skipped (if local).
+        @param trusted: Whether validation should be skipped (if local).
         """
         sources_file_path = self._get_internal_sources_list()
         source_options = ""
-        if trusted is not None:
+        if trusted is not None and url.startswith("file:"):
             trusted_val = "yes" if trusted else "no"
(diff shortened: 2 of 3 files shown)


simpoir:
If you test through py2->py3 upgrade, you'll get issues until https://github.com/CanonicalLtd/landscape-client/pull/30 is merged

landscape/lib/apt/package/facade.py
```diff
@@ -269,15 +269,21 @@ def _get_internal_sources_list(self):
...
 
-    def add_channel_apt_deb(self, url, codename, components=None):
+    def add_channel_apt_deb(self, url, codename, components=None,
+                            trusted=None):
         """Add a deb URL which points to a repository.
 
         @param url: The base URL of the repository.
         @param codename: The dist in the repository.
         @param components: The components to be included.
+        @param trusted: Whether validation shoule be skipped (if local).
```
blackboxsw: s/shoule/should/

debian/landscape-client.postinst
```diff
@@ -106,6 +106,15 @@ END
...
         install --owner=landscape /dev/null $USER_UPDATE_FLAG_FILE
         echo "This file indicates that the Landscape client needs to send updated user information to the server." >> $USER_UPDATE_FLAG_FILE
+
+        # To work around bug #1735100 we rewrite file-local landscape sources
+        # with the trusted flag, as they have no release file, thus are
+        # unsigned repositories.
+        LANDSCAPE_INTERNAL_SOURCES=/etc/apt/sources.list.d/_landscape-internal-facade.list
+        if grep -q -e "^deb file:" $LANDSCAPE_INTERNAL_SOURCES; then
+            cp $LANDSCAPE_INTERNAL_SOURCES $LANDSCAPE_INTERNAL_SOURCES.orig
+            sed -i 's/^deb file:/deb [ trusted=yes ] file:/' $LANDSCAPE_INTERNAL_SOURCES
```
blackboxsw: Hrm, so do we also need to kick apt-get update here too now that we've re-written the sources file?

landscape/lib/apt/package/facade.py
```diff
@@ -269,15 +269,21 @@ def _get_internal_sources_list(self):
...
         """
         sources_file_path = self._get_internal_sources_list()
-        sources_line = "deb %s %s" % (url, codename)
+        source_options = ""
+        if trusted is not None:
+            trusted_val = "yes" if trusted else "no"
+            source_options = "[ trusted={} ] ".format(trusted_val)
```
blackboxsw: Could collapse these two lines:
if trusted is not None:
        source_options = "[ trusted={} ] ".format("yes" if trusted else "no")

landscape/lib/apt/package/facade.py
```diff
@@ -269,15 +269,21 @@ def _get_internal_sources_list(self):
...
 
         @param url: The base URL of the repository.
         @param codename: The dist in the repository.
         @param components: The components to be included.
+        @param trusted: Whether validation shoule be skipped (if local).
         """
         sources_file_path = self._get_internal_sources_list()
-        sources_line = "deb %s %s" % (url, codename)
+        source_options = ""
+        if trusted is not None:
```
blackboxsw: We probably should also confirm that url.startswith('file:')  before trying to add trusted=yes right?   Even though it looks like all the calls are called with file://%s

landscape/lib/apt/package/facade.py
blackboxsw: Can we also get an unit test on facade to validate trusted=yes|no behavior.

debian/landscape-client.postinst
simpoir: landscape-client's package-monitor should take care of that once the source entry is correct. Also triggering apt from a package update triggered by apt seems like a bad plan.

landscape/lib/apt/package/facade.py
simpoir: good catch

landscape/lib/apt/package/facade.py
simpoir: That would bust the 80 char and force me to add a line break… Unless you insist I think it's clean as it is.

landscape/lib/apt/package/facade.py
simpoir: Although the only callsite is add_channel_deb_dir() which is a file:// URL, I think it would be a safe check to do as that flag could be easily misused. Consider it done.

debian/landscape-client.postinst
```diff
@@ -106,6 +106,15 @@ END
         USER_UPDATE_FLAG_FILE="$DATA_PATH/user-update-flag"
         install --owner=landscape /dev/null $USER_UPDATE_FLAG_FILE
         echo "This file indicates that the Landscape client needs to send updated user information to the server." >> $USER_UPDATE_FLAG_FILE
+
+        # To work around bug #1735100 we rewrite file-local landscape sources
+        # with the trusted flag, as they have no release file, thus are
+        # unsigned repositories.
+        LANDSCAPE_INTERNAL_SOURCES=/etc/apt/sources.list.d/_landscape-internal-facade.list
+        if grep -q -e "^deb file:" $LANDSCAPE_INTERNAL_SOURCES; then
+            cp $LANDSCAPE_INTERNAL_SOURCES $LANDSCAPE_INTERNAL_SOURCES.orig
```
blackboxsw: It feels strange to leave orig this around even after we've successfully completed the package profile install. We probably should remove it when we LANDSACPE_INTERNAL_SOURCES (on success)

simpoir:
agreed. Knowing that the internal source file should be removed after the operation, I don't see the need to keep a backup. I'll remove it

PR #32, opened 2018-02-03
Replace usermod calls by PAM to remove passlib (LP: #1743558)

simpoir:
Testing instructions:

* build and install package
  * check package does not depend on python3-passlib (universe)
* register client
* go to computer/users tab
  * create a new user
  * edit an existing user password
  * edit root password
* verify passwords changed correctly


simpoir:
Side note: we could just replace the hash library by another which would be supported, however hashing with a weak algorithm (salted MD5) or passing password entries to usermod are both bad strategies; the algorithm is considered weak and deprecated; the usermod argument is insecure (see man page)

panlinux:
Hi,

I was working on using this diff (works, just doesn't have tests):
```diff
diff --git a/landscape/client/user/management.py b/landscape/client/user/management.py
index e618f00f..8c1f4f11 100644
--- a/landscape/client/user/management.py
+++ b/landscape/client/user/management.py
@@ -48,10 +48,15 @@ class UserManagement(object):
         return output
 
     def _set_password(self, username, password):
-        password = password.encode("utf-8")
-        crypted = md5_crypt.encrypt(password)
-        result, output = self.call_popen(["usermod", "-p", crypted, username])
+        chpasswd_input = "{}:{}\n".format(username, password).encode("utf-8")
+        chpasswd = self._provider.popen(["chpasswd"],
+                                        stdin=subprocess.PIPE,
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.STDOUT)
+        output, _ = chpasswd.communicate(chpasswd_input)
+        result = chpasswd.returncode
         if result != 0:
+            username = username.encode("utf-8")
             raise UserManagementError("Error setting password for user "
                                       "%s.\n%s" % (username, output))
         return output
```
And then I saw that this bug isn't so urgent anymore, because python-passlib (the py2 version) is in main already. I commented as such in #1743558 and changed its priority.

That being said, there is still an argument to not hash the password ourselves, specially with a broken hash like md5. But I still think the chpasswd approach above is simpler and it will use whatever the system has configured locally, which is pam most of the time.

simpoir:
I'll resubmit this with @panlinux 's patch and tests.

PR #34, opened 2018-02-07
decode message types in task handler and continue on unknown type

simpoir:
Testing:

* install landscape-client on artful/bionic
  * kick a package profile (should break because of apt-secure being default in distro, but the task will stay)
* upgrade
* verify package-changer/monitor logs for stack trace


panlinux:
What is "APT issue in distro"?

PR #40, opened 2018-03-01
changelog update

panlinux:
d/changelog update from the bionic upload, and opening up 18.03 development. Doesn't have to be 18.03, but it must be higher than 18.01 because that's in bionic now.


panlinux:
Maybe I should also leave the release as "UNRELEASED" instead of bionic. Seems to be working before. I don't mind either way.

PR #42, opened 2018-03-07
Remove landscape from motd take 2

panlinux:
This just removes `LandscapeLink` from the list of default plugins. It's much nicer for the upgrade case because then we don't have to worry about users who have explicitly decided to enable this plugin, because it still works. If we remove the plugin and hit such a case, then `landscape-sysinfo` crashes (see https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/1754002).

Thanks @simpoir for the suggestion :)


panlinux:
https://github.com/CanonicalLtd/landscape-client/pull/43 addresses that, @sparkiegeek . Thanks

PR #45, opened 2018-04-05
Override the apt progress pulse().

simpoir:
This should remove the tracebacks under python3 + bionic. (LP: #1758529)

Testing instructions:

```
edit root-client.conf
sudo ./scripts/landscape-client -r root-client.conf
log to landscape, accept the client, wait for packages to sync.
search for a package (e.g. fortune), mark it for install; wait.
grep pulse /tmp/landscape-root/*.log  # should return nothing
```


simpoir:
FWIW, I've marked the bug as affecting python-apt. Whether it gets fixed or not remains to be seen, but from our standpoint, we should be ok.

PR #48, opened 2018-05-17
Handle unicode apt-sources-replace (LP: #1771443)

simpoir:
Support both unicode and bytes content for apt-sources-replace,
as content should be unicode, but can be passed as bytes after a client
upgrade from python 2.7

Testing instructions:

* on a bionic container, ./scripts/landscape-client -c root-client.conf
* on the server,
  ```
make freshdata && ./dev/racecar
  source ./dev/setup_api
  curl 'https://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search=0xC0B21F32' > mirror_key
  landscape-api import-gpg-key ubuntu-new-mirror-key ./mirror_key
  // the sign-key comes from freshdata
  landscape-api create-series --pockets security --components restricted --architectures amd64 --mirror-gpg-key ubuntu-new-mirror-key --gpg-key sign-key --mirror-uri http://archive.ubuntu.com/ubuntu/ --mirror-series bionic bionic-repo ubuntu
  landscape-api sync-mirror-pocket security bionic-repo ubuntu

  // create a profile
  landscape-api create-repository-profile --description "This profile is for Bionic-security-restricted" bio-secu-restricted-profile
  landscape-api associate-repository-profile --tags bionic-test bio-secu-restricted-profile

  // accept the bionic container, set the 'bionic-test' tag
  // the profile should apply, wait for it to succeed
  landscape-api add-pockets-to-repository-profile bio-secu-restricted-profile security bionic-repo ubuntu
  // the profile should apply again, wait for it to succeed again

```

### DIFF
--- a/landscape/client/manager/aptsources.py
+++ b/landscape/client/manager/aptsources.py
@@ -128,7 +128,10 @@ class AptSources(ManagerPlugin):
         for source in sources:
             filename = os.path.join(self.SOURCES_LIST_D,
                                     "landscape-%s.list" % source["name"])
-            with open(filename, "wb") as sources_file:
+            # Servers send unicode, but an upgrade from python2 can get bytes
+            # from stored messages, so we need to handle both.
+            is_unicode = isinstance(source["content"], type(u""))
+            with open(filename, ("w" if is_unicode else "wb")) as sources_file:
                 sources_file.write(source["content"])
             os.chmod(filename, 0o644)
         return self._run_reporter().addCallback(lambda ignored: None)
--- a/landscape/client/manager/tests/test_aptsources.py
+++ b/landscape/client/manager/tests/test_aptsources.py
@@ -105,6 +105,23 @@ class AptSourcesTests(LandscapeTest):
         with open(saved_sources_path) as saved_sources:
             self.assertEqual("original content\n", saved_sources.read())
 
+    def test_sources_list_unicode(self):
+        """
+        When receiving apt-sources-replace, client correctly also handles
+        unicode content correctly.
+        """
+        self.manager.dispatch_message(
+            {"type": "apt-sources-replace",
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by simpoir** [comment]
[18.04 Repository Profiles - fail to apply](https://trello.com/c/fEcKsJWo/683-1804-repository-profiles-fail-to-apply)

File: landscape/client/manager/aptsources.py [hunk 1]
```diff
@@ -128,7 +128,10 @@ def _handle_sources(self, ignored, sources):
         for source in sources:
             filename = os.path.join(self.SOURCES_LIST_D,
                                     "landscape-%s.list" % source["name"])
-            with open(filename, "wb") as sources_file:
+            # Servers send unicode, but an upgrade from python2 can get bytes
+            # from stored messages, so we need to handle both.
+            is_unicode = isinstance(source["content"], type(u""))
+            with open(filename, ("w" if is_unicode else "wb")) as sources_file:
```
waveform80: I *want* to rail against this and say "if it's a text file, we should be writing it as a text file", i.e. `with io.open(filename, 'w', encoding=ENCODING) as f: f.write(source['content'].decode(ENCODING))` ... the problem is ENCODING is essentially undefined (deb822 defines it as UTF-8 but that doesn't apply to the older .list style, and besides we don't explicitly know the encoding of `source['content']` either). So, this is probably the best we can do for now.

simpoir: The problem isn't so much the encoding (files are probably ascii as they contain only urls written by landscape server). The issue comes from str getting interpreted as unicode for py3 and bytes for py2, which affected how some of our libraries. I wish we would have taken a slightly different strategy when porting to python3 (which would have highlighted more of those inconsistencies earlier). Unfortunately, this kind of patch is the one with the least impact on what is already released.

PR #51, opened 2018-07-25
Permit environments containing unicode chars for script execution

waveform80:
The server supplied environment for a client-executed script may include unicode strings (e.g. when a computer's tag includes diacriticals). Currently, twisted throws an exception when encountering these as the default encoding it uses (on Python 2) is 'ascii'. This patch ensures all unicode strings in the environment are encoded appropriately prior to being passed to twisted for execution. The encoding used is from `sys.getfilesystemencoding()` as this is what Python 3 uses for its `os.environ` dict.

Furthermore, it ensures that various locale-relevant vars (LC_ALL, LC_CTYPE, LANG) aren't stripped from the environment, permitting the executed script to know what encoding to expect when querying the environment.

### DIFF
--- a/landscape/client/manager/scriptexecution.py
+++ b/landscape/client/manager/scriptexecution.py
@@ -4,6 +4,7 @@ Functionality for running arbitrary shell scripts.
 @var ALL_USERS: A token indicating all users should be allowed.
 """
 import os
+import sys
 import os.path
 import tempfile
 import shutil
@@ -105,6 +106,11 @@ class ScriptRunnerMixin(object):
             uid = None
         if gid == os.getgid():
             gid = None
+        env = {
+            key: value.encode(sys.getfilesystemencoding(), errors='replace')
+            if isinstance(value, unicode) else value
+            for key, value in env.items()
+        }
 
         pp = ProcessAccumulationProtocol(
             self.registry.reactor, self.size_limit, self.truncation_indicator)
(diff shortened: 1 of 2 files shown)


landscape/client/manager/scriptexecution.py
```diff
@@ -105,6 +106,11 @@ def _run_script(self, filename, uid, gid, path, env, time_limit):
             uid = None
         if gid == os.getgid():
             gid = None
+        env = {
+            key: value.encode(sys.getfilesystemencoding(), errors='replace')
```
simpoir: Shouldn't this be using sys.getdefaultencoding()? Filesystem encoding is could similar but that doesn't sound right.

landscape/client/manager/scriptexecution.py
```diff
@@ -105,6 +106,11 @@ def _run_script(self, filename, uid, gid, path, env, time_limit):
...
+            if isinstance(value, unicode) else value
```
simpoir: nitpick: "ifs" in list comprehension value reads weird. Parentheses and indentation would help.

landscape/client/manager/tests/test_scriptexecution.py
```diff
@@ -99,6 +110,35 @@ def check_environment(results):
         result.addCallback(check_environment)
         return result
 
+    def test_server_supplied_env_with_funky_chars(self):
+        """
+        Server-supplied environment variables can be unicode strings; client
+        should pass these to the script's environment encoded appropriately
+        (encoding from Python's sys.getfilesystemencoding).
```
simpoir: Again, I think it should be sys.getdefaultencoding.

landscape/client/manager/scriptexecution.py
waveform80: That's what I originally thought, but Python 3 uses `sys.getfilesystemencoding` for its [os.environ implementation](https://docs.python.org/3/library/os.html#os.environ) (the unencoded bytes-variant of the environment is os.environb in Python 3). I guess it makes a certain amount of sense as the most important vars in the environment are usually file-system related (PATH etc.). There's also the aspect that `sys.getdefaultencoding()` on Python 2 simply returns 'ascii' regardless of the environment.

landscape/client/manager/scriptexecution.py
waveform80: Fair enough - will push an update

landscape/client/manager/scriptexecution.py
simpoir: fair enough. just drop a comment for future me to remember that. :)

PR #52, opened 2018-08-16
check chassis_vendor first for vm_info to allow for customized sys_vendor

asciiprod:
Hetzner Cloud uses dmi sys_vendor to identify itself (e.g. to cloud-init).
This confuses landscape, so it needs to be treated like KVM

### DIFF
--- a/landscape/lib/tests/test_vm_info.py
+++ b/landscape/lib/tests/test_vm_info.py
@@ -79,6 +79,13 @@ class GetVMInfoTest(BaseTestCase):
         self.make_dmi_info("sys_vendor", "DigitalOcean")
         self.assertEqual(b"kvm", get_vm_info(root_path=self.root_path))
 
+    def test_get_vm_info_with_hetzner_sys_vendor(self):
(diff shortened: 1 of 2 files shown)


asciiprod:
I agree that chassis_vendor is probably a better approach. Our newer QEMU/Seabios will have 'QEMU' in /sys/class/dmi/id/chassis_vendor while DO still has 'Bochs'

asciiprod:
I am not aware of anyone else setting chassis to QEMU. So this should be much better to detect KVM based servers than the previous order. If I am not mistaken, after merging this change, you could remove Openstack and DO. I haven't checked AWS, but possibly that as well.

simpoir:
Thanks again for your contribution, but I'm closing this PR; the proposed change does not appear to fix any currently known issue and is unlikely to be released.

Unless there is a known issue with the detection which would warrant an SRU, this won't go through. For future releases of the client we will want to rely on the (smarter) systemd-detect-virt instead of the mapping we're using here.

PR #55, opened 2018-08-23
Reboot means reboot! (LP: #1670291)

waveform80:
Add an additional step when shutdown fails to power-off or restart the machine; call the /sbin/poweroff or /sbin/reboot binaries instead to attempt an immediate shutdown. We can't (reliably) report success or failure from this, so the operation will still appear failed but it appears more reliable practice runs with VMs and containers in post-trusty upgrade environments. (LP: #1670291)

## Testing instructions

* Launch a trusty VM and/or container
* Patch landscape-client in the instance
* Register with landscape
* Request an upgrade to xenial
* Upgrade should complete successfully; don't reboot the instance yet!
* Re-patch the landscape-client (because the release-upgrade will have over-written the prior patch) and restart the landscape-client
* Request a restart of the machine from within landscape
* Restart should *appear* to fail (from landscape's perspective) but should ultimately succeed on the instance
* An overly wordy message should appear in the failure log explaining what's going on :)
* All tests should still pass

### DIFF
--- a/landscape/client/manager/shutdownmanager.py
+++ b/landscape/client/manager/shutdownmanager.py
@@ -63,12 +63,12 @@ class ShutdownManager(ManagerPlugin):
         failure_report = '\n'.join([
             failure.value.data,
             "",
-            "Attempting to force %(operation)s. Please note that if this "
+            "Attempting to force {operation}. Please note that if this "
             "succeeds, Landscape will have no way of knowing and will still "
             "mark this activity as having failed. It is recommended you check "
             "the state of the machine manually to determine whether "
-            "%(operation)s succeeded." %
-            {"operation": "reboot" if reboot else "shutdown"}
+            "{operation} succeeded.".format(
+                operation="reboot" if reboot else "shutdown")
         ])
         deferred = self._respond(FAILED, failure_report, operation_id)
         # Add another callback spawning the poweroff or reboot command (which
--- a/landscape/client/manager/shutdownmanager.py
+++ b/landscape/client/manager/shutdownmanager.py
@@ -45,7 +45,7 @@ class ShutdownManager(ManagerPlugin):
         protocol = ShutdownProcessProtocol()
         protocol.set_timeout(self.registry.reactor)
         protocol.result.addCallback(self._respond_success, operation_id)
-        protocol.result.addErrback(self._respond_failure, operation_id)
+        protocol.result.addErrback(self._respond_failure, operation_id, reboot)
         command, args = self._get_command_and_args(protocol, reboot)
         self._process_factory.spawnProcess(protocol, command, args=args)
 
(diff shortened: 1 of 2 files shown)


landscape/client/manager/shutdownmanager.py
```diff
@@ -58,9 +58,31 @@ def _respond_success(self, data, operation_id):
...
+        failure_report = '\n'.join([
+            failure.value.data,
+            "",
+            "Attempting to force %(operation)s. Please note that if this "
+            "succeeds, Landscape will have no way of knowing and will still "
+            "mark this activity as having failed. It is recommended you check "
+            "the state of the machine manually to determine whether "
+            "%(operation)s succeeded." %
+            {"operation": "reboot" if reboot else "shutdown"}
+        ])
```
simpoir: IMHO, a "{}\n\nBlah blah {0}...".format(operation=...)
would be mildly more readable than this join and % format, but feel free to disregard my opinion.

landscape/client/manager/shutdownmanager.py
```diff
@@ -58,9 +58,31 @@ def _respond_success(self, data, operation_id):
...
+        deferred = self._respond(FAILED, failure_report, operation_id)
+        # Add another callback spawning the poweroff or reboot command (which
+        # seem more reliable in aberrant situations like a post-trusty release
+        # upgrade where upstart has been replaced with systemd). If this
+        # succeeds, we won't have any opportunity to report it and if it fails
+        # we'll already have responded indicating we're attempting to force
+        # the operation so either way there's no sense capturing output
```
simpoir: I disagree: a log entry on failure would be better than dropping the input. But yeah, on success, it goes down too fast to handle any result.

landscape/client/manager/shutdownmanager.py
```diff
@@ -58,9 +58,31 @@ def _respond_success(self, data, operation_id):
...
+        protocol = ProcessProtocol()
+        command, args = self._get_command_and_args(protocol, reboot, True)
+        deferred.addCallback(
+            lambda _: self._process_factory.spawnProcess(
+                protocol, command, args=args))
+        return deferred
```
simpoir: I like that you still report it as a failure (it was). It's weird if it did succed, but there was an anomaly worth checking and the message is pretty clear.

landscape/client/manager/shutdownmanager.py
waveform80: Sure, will replace with .format

landscape/client/manager/shutdownmanager.py
waveform80: Yeah, that's sort of what I'm alluding to in that comment; if we delay reporting the status until we actually try reboot / poweroff then in the case of success we won't have any opportunity to report it, in which case we'd successfully report failure and fail to report success :) Hence, I think the current situation is preferable: always report, but warn in the case of failure that it might actually be success.

PR #56, opened 2018-10-02
Clear hash id database on package resync.

simpoir:
This should avoid loop when inconsistencies are present in the database
through either corruption or server-restore. (LP: #1616116)

Testing instructions:

* launch a landscape server and register (using ppa packages is somewhat simpler as hash-db is pre-populated).
* register client against server. wait for package info
* pg_dumpall
* add a repo and wait for new package to show on client.
* restore postgres backup.
* trigger a package install from the new repo to have new package show up
* if you haven't regenerated hash-ids on the server, clients will be stuck in a loop. this is to be expected as the server is inconsistent with itself.
* client should resync once then will re-fetch hash on the next run. Some hash-id requests may go through in the meantime.

### DIFF
--- a/landscape/client/package/tests/test_reporter.py
+++ b/landscape/client/package/tests/test_reporter.py
@@ -1204,10 +1204,11 @@ class PackageReporterAptTest(LandscapeTest):
 
         self.assertEqual(expected, command)
 
+    @inlineCallbacks
     def test_resynchronize(self):
         """
         When a resynchronize task arrives, the reporter should clear
-        out all the data in the package store, except the hash ids.
+        out all the data in the package store, including the hash ids.
 
         This is done in the reporter so that we know it happens when
         no other reporter is possibly running at the same time.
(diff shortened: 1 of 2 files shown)


simpoir:
A nice side-effect of dropping the hash database is a delay is induced in the package info refresh, thus throttling the refresh and avoiding the server being overloaded.

landscape/client/package/tests/test_reporter.py
```diff
@@ -1243,30 +1255,35 @@ def test_resynchronize(self):
...
+
+        # The hashes should go away to avoid loops
+        # when server gets restored to an early state.
+        hash1 = self.store.get_hash_id(foo_hash)
+        hash2 = self.store.get_hash_id(HASH2)
+        self.assertNotEqual(hash1, 3)
+        self.assertNotEqual(hash2, 4)
+        self.assertFalse(os.path.exists(hash_id_file))
+
+        # But the other data should.
```
sparkiegeek: This isn't a 'but ... should' anymore since both of them go away :smile:

landscape/client/package/tests/test_reporter.py
```diff
@@ -1220,6 +1221,17 @@ def test_resynchronize(self):
...
+
+        # create hash_id file
+        message_store.set_server_uuid("uuid")
+        self.facade.set_arch("arch")
+        hash_id_file = os.path.join(
+            self.config.hash_id_directory, "uuid_codename_arch")
+        os.makedirs(self.config.hash_id_directory)
+        with open(hash_id_file, "w") as f:
+            pass
+        self.assertFileContent(hash_id_file, b"")
```
sparkiegeek: this assert seems a bit redundant, don't think we need to test behaviour of `open(..., 'w')`

PR #57, opened 2018-10-15
report packages from security-pocket

simpoir:
Testing instructions:

sudo ./scripts/landscape-client -c path_to_your_favorite_config

Check the debug logs for packages ids being reported. in addition to available,
there should be a couple of security.

### DIFF
--- a/landscape/client/package/reporter.py
+++ b/landscape/client/package/reporter.py
@@ -609,6 +609,9 @@ class PackageReporter(PackageTaskHandler):
         - are now set, and were not;
         - were previously set but are not anymore;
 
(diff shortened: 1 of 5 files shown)


landscape/client/package/reporter.py
```diff
@@ -664,6 +670,13 @@ def _compute_packages_changes(self):
                 if self._facade.is_package_upgrade(package):
                     current_upgrades.add(id)
 
+                # Is this package present in the security pocket?
+                security_origins = [
```
sparkiegeek: given you're treating this as a boolean, you could escape early by using `any()`

PR #58, opened 2018-10-29
Allow bad chars in system vendor files (LP: #1799843)

waveform80:
And while we're at it, clean up some of the mess in fs.py: there's
almost no circumstances where we want landscape client to fall over
because a text file has invalid UTF-8 (the original function before the
py2/3 split had no such circumstances that'd lead to decoding errors).
Furthermore, read_binary_file's limit was wrong when positive (not that
anything called it with a positive limit) and read_text_file defeated
the purpose of having a limit by permitting an arbitrarily large read
even when one was set.

### DIFF
--- a/landscape/lib/tests/test_fs.py
+++ b/landscape/lib/tests/test_fs.py
@@ -28,11 +28,11 @@ class ReadFileTest(BaseTestCase):
 
     def test_read_binary_file_with_limit(self):
         """
-        With a positive limit L{read_binary_file} reads only the bytes after
-        the given limit.
+        With a positive limit L{read_binary_file} reads up to L{limit} bytes
+        from the start of the file.
         """
         path = self.makeFile("foo bar")
-        self.assertEqual(read_binary_file(path, limit=3), b" bar")
+        self.assertEqual(read_binary_file(path, limit=3), b"foo")
 
     def test_read_binary_file_with_negative_limit(self):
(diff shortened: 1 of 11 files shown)


landscape/lib/tests/test_fs.py
```diff
@@ -64,12 +64,12 @@ def test_read_text_file(self):
...
         """
-        With a positive limit L{read_text_file} returns only the characters
-        after the first L{limit} characters as string.
+        With a positive limit L{read_text_file} returns up to L{limit}
+        characters from the start of the file.
         """
         utf8_content = codecs.encode(u"foo \N{SNOWMAN}", "utf-8")
         path = self.makeFile(utf8_content, mode="wb")
-        self.assertEqual(read_text_file(path, limit=3), u" ☃")
+        self.assertEqual(read_text_file(path, limit=3), u"foo")
```
sparkiegeek: It'd be nice to have a test for a string with characters that encode to > 1 UTF-8 byte, and assert the correct behaviour of the decode with replace

waveform80:
Good point, I'll bung one in

landscape/lib/tests/test_fs.py
waveform80: Resolved in ad801b2c93e560685d1722ed76375975688534d0

PR #63, opened 2019-02-21
Add option to stagger launch of broker plugins.

simpoir:
Testing instructions:
sudo ./scripts/landscape-config --stagger-launch 0.5 -n -c root-client.conf
cat root-client.conf | grep stagger
sudo ./scripts/landscape-client --log-level debug -c root-client.conf
Check for debug log entries mentioning monitors getting delayed


sparkiegeek:
Why do we want this as an option? I think this should be the default behaviour, and always enabled.

simpoir:
I'm fine with making delayed start the default behaviour. I was thinking that, for SRU-purposes, it might be better to keep the existing behaviour by default, but the added delay should be fairly safe anyway.

I'd still keep it as an option, if only to remove delays when running tests.

4a6f656c:
Rather than just skewing the start time, could you please skew each run (including the start) by some interval (normally +/- 10%) - this avoids a range of synchronization issues and avoids having colliding periodic loads on a machine (for example, consider a cron job set to run at /12 and /42 - only skewing the start interval means that if your start ends up running at /12, you're now going to collide and run at exactly the same time as that cron job on every run).

By skewing on every interval, on average you will still run the same number of runs within the same extended time frame, however any synchronisation with other jobs (or clients) will only be temporary, since they are almost guaranteed to desynchronise on the very next run.

simpoir:
I like the idea of an additional delay, but I'll submit that as a separate branch.

PR #65, opened 2019-05-14
Converts the sysvinit service to a systemd service

simpoir:
This branch adds the systemd service and  migrates service config to postinst.

Testing instructions:

dpkg-buildpackage -b && dpkg -i ../landscape-c*.deb && landscape-config
I suggest trying a clean install, an upgrade from an unconfigured client, an upgrade from a configured client.


debian/landscape-client.postinst
```diff
@@ -116,6 +116,26 @@ END
...
+            update-rc.d landscape-client remove >/dev/null
+            rm -f /etc/init.d/landscape-client
+        fi
+
+        # Migrate old sysvinit config to systemd
+        DEFAULTS=/etc/default/landscape-client
+        if [ -f $DEFAULTS ]; then
+            RUN=$(. $DEFAULTS; echo $RUN)
+            if [ "$RUN" = "1" ]; then
+                systemctl enable landscape-client.service
```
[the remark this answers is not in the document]
simpoir: I don't think that's needed, as dh_systemd_start should start it automatically below if the service is enabled.

landscape/client/sysvconfig.py
```diff
@@ -1,60 +1,41 @@
...
     """ Error running a process with os.system. """
 
 
-class SysVConfig(object):
-    """Configure and drive the Landscape client init script.
+class SystemdConfig(object):
+    """Configure and drive the Landscape client service.
 
-    @param filename: Path to the file holding init scripts env variables.
+    @param filename: Path to the file holding service env variables.
```
sparkiegeek: remove the lie

debian/landscape-client.service
```diff
@@ -0,0 +1,15 @@
...
+After=local-fs.target
+After=remote-fs.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=forking
+Group=landscape
+ExecStart=/usr/bin/landscape-client --daemon --pid-file /run/landscape/landscape-client.pid
```
sparkiegeek: hmm, won't we be better off not forking (i.e. no `--daemon`)and having `systemd` invoke `landscape-client` directly?

simpoir: It doesn't make a lot of difference here. Everything is ultimately tracked with cgroups.

I'd have to test a bit more if we change to a simple service to make sure everything runs nice from all the entry points as I have some doubts.

simpoir: correction. I've changed to simple service as suggested. It detects better if the service fails to start for whatever reason.

simpoir:
@maxiberta I've changed the service to a "simple" type. I think it solves the update hang you had on invalid config, amongst other things.

PR #69, opened 2019-09-04
Use sysfs instead of deprecated procfs IF for thermal, if available.

borospeti:


landscape/lib/sysstats.py
```diff
@@ -104,17 +108,27 @@ class ThermalZone(object):
...
-                    self.temperature = line[12:].strip()
-                    try:
-                        value, unit = self.temperature.split()
-                        self.temperature_value = int(value)
-                        self.temperature_unit = unit
-                    except ValueError:
-                        pass
+            with open(temperature_path) as f:
+                line = f.readline()
+                if re.match(r'^\d+\s*$', line):
```
maxiberta: I'd rather use a `try: except` block around `int()` as in the `else:` codepath as the most common, non-breaking case should be much faster.

landscape/lib/sysstats.py
```diff
@@ -104,17 +108,27 @@ class ThermalZone(object):
...
+                    self.temperature_value = int(line.strip()) / 1000
+                    self.temperature_unit = 'C'
+                    self.temperature = '{:.1f}{}'.format(
```
maxiberta: This is just a nitpick, but based on tests in `landscape/lib/tests/test_sysstats.py` I think there should be a whitespace between value and unit.

landscape/lib/sysstats.py
borospeti: Good catch, fixed.

landscape/lib/sysstats.py
borospeti: Fixed.

PR #71, opened 2019-10-24
[debian] Keeping src pkg up-to-date

slashdd:
Moving from compat v7 (deprecated) to v12
 - W: package-uses-deprecated-debhelper-compat-version 7

Fix lintian Warning report:
 - W: ancient-python-version-field x-python3-version 3.5
 - W: ancient-standards-version 3.8.2 (released 2009-06-16) (current is 4.4.0)

Fix lintian Informational/Pedantic details:
 - I: debian-pycompat-is-obsolete
 - P: no-homepage-field

Signed-off-by: Eric Desrochers <eric.desrochers@canonical.com>


debian/rules
```diff
@@ -1,103 +1,18 @@
...
-#
-binary-arch: build install
-	dh_lintian
-	dh_testdir
-	dh_testroot
-	dh_installdocs
-	dh_installman
-	dh_installchangelogs
-	dh_install --sourcedir debian/tmp/
-	dh_systemd_enable --no-enable
```
simpoir: I'm not sure dh does what we want by default here. We used to not enable the service by default; only when landscape-config wrote the client configuration. We probably want to keep that to avoid the unit failing after install.

slashdd: The only valuable item to keep here is the `dh_systemd_enable --no-enable`

I'll push an updated commit to include it.

debian/rules
```diff
@@ -1,103 +1,21 @@
...
-#
-binary-arch: build install
-	dh_lintian
-	dh_testdir
-	dh_testroot
-	dh_installdocs
-	dh_installman
-	dh_installchangelogs
-	dh_install --sourcedir debian/tmp/
+override_dh_systemd_enable:
```
simpoir: This change doesn't seem to have the desired effect (disabled/stopped by default). I don't think dh_systemd_enable is even used if I look at the resulting postinst script. However, this did, whether idiomatic or not, create a disabled service which only gets (re)started if landscape-config enabled it:

override_dh_installsystemd:                                                     
  dh_installsystemd --no-enable --no-start

slashdd: You are right, good catch here:

## dh_systemd_enable (from debhelper 12)
```
if (not compat(10)) {
        error("dh_systemd_enable is no longer used in compat >= 11, please use dh_installsystemd instead");
```

slashdd: Fixed in my latest push

debian/rules
```diff
@@ -1,103 +1,21 @@
...
-build-arch: build
+%:
+	dh $@ --with python3 --buildsystem=pybuild
 
-build-indep: build
-
-build: build-stamp
-build-stamp:
-	dh_testdir
-	sed -i -e "s/^DEBIAN_REVISION = \"\"/DEBIAN_REVISION = \"-$(revision)\"/g" landscape/__init__.py
```
simpoir: That version is used as user-agent version to identify regressions from the server side. I think we should keep it. Not sure how's the proper way do to it, but if I had to, 
```
override_dh_auto_build:
       sed -i -e "s/^DEBIAN_REVISION = \"\"/DEBIAN_REVISION = \"-$(shell dpkg-parsechangelog | sed -rne 's,^Version: ([^-]*)-(.*),\2, p')\"/g" landscape/__init__.py
       dh_auto_build
```

PR #73, opened 2019-11-27
Sort Packages file in tests.

simpoir:
This change should cover a few failures we've seen with recent apt (eoan, focal). Those were due to the test Packages file getting sorted mid-test thus making package ID change and breaking the test.

Basically, I just added sorting when writing Packages in test fixtures.

### DIFF
--- a/landscape/lib/apt/package/testing.py
+++ b/landscape/lib/apt/package/testing.py
@@ -1,5 +1,6 @@
 import base64
 import os
+import textwrap
 import time
 
 import apt_inst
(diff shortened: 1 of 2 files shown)


landscape/lib/apt/package/testing.py
```diff
@@ -1,5 +1,6 @@
 import base64
 import os
+import textwrap
```
maxiberta: Looks like this module is not used?

maxiberta:
Oh, this was closed! Np, most probably superseded by #74 .

PR #77, opened 2019-12-07
Handle EINVAL error of SIOCETHTOOL ioctl

rbalint:
The error occurs on WSL (1) and causes an error showing up in MOTD.

LP: #1855522


rbalint:
How can I help getting this merged? I'd like to update the package with this fix and SRU it to Eoan in the next days.

PR #81, opened 2020-04-07
Clean up stale lock files

simpoir:
This change tries to clean up stale lock files, which points to dead processes. (LP: #1870087)

Testing instructions:

install/register client.
```
# systemctl stop landscape-client
# cd /var/lib/landscape/client/sockets/
# ln -sf 1 broker.sock.lock # make some fake locks pointing to live processes (1 is nice because it's always there and landscape should not have access to it)
# ln -sf 1 manager.sock.lock
# ln -sf 1 monitor.sock.lock
# systemctl start landscape-client
# ls -l cd /var/lib/landscape/client/sockets/
```
your fake locks should have been replaced by and point to valid PIDs (not 1).
You might as well check that lock conflicts do conflict while services are running:
```
# landscape-broker
twisted.internet.error.CannotListenError: Couldn't listen on any:b'/var/lib/landscape/client/sockets/broker.sock
```

### DIFF
--- a/landscape/client/lockfile.py
+++ b/landscape/client/lockfile.py
@@ -1,4 +1,3 @@
-import errno
 import os
 
 from twisted.python import lockfile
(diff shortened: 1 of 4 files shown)


simpoir:
I now raise the lock as unclean when we detect a stale lock.
Previous would fail, then re-clean itself now that the lock pid was definitely dead.
This should take care of that exception without the extra pass.

PR #84, opened 2020-05-04
Add landscape-sysinfo CLI arg to specify maximum column width

pirate:
Parsing the output of `landscape-sysinfo` is difficult across multiple system right now, as some systems will print output in two columns and others in one.  The simplest solution I could find to force it to always output 1 column is to pass a very low value to `format_sysinfo(..., width=1)` so that it always prints in one column.


juanmitaboada:
I have created a new PR #156 which corrects this PR

PR #86, opened 2020-07-17
Add RHEV to vm_info map. (LP: #1884116)

simpoir:
Testing instructions:

The easy way to verify this is to run the unit tests.
The proper way to verify this is to spawn a kvm and `virsh edit`

and add a `<smbios mode="sysinfo"/>` under `<os>`, and a sysinfo block after `</os>`:
```
<sysinfo type='smbios'>
  <bios>
    <entry name='vendor'>LENOVO</entry>
  </bios>
  <system>
    <entry name='manufacturer'>Red Hat</entry>
    <entry name='product'>RHEV Hypervisor</entry>
    <entry name='version'>7.8-2.el7ev</entry>
  </system>
  <chassis>
    <entry name='manufacturer'>Red Hat</entry>
  </chassis>
</sysinfo>
```

The other way is to have an RHEV install.


maxiberta (approved):
LGTM

PR #88, opened 2020-08-12
Return missing deferred on service shutdown.

simpoir:
This should allow services to let their publisher shut down and clean
after itself, thus avoiding stale locks and sockets. (LP: #1870087)

### Testing instructions:

Spawn and stop the client. There should not be any lock or socket left.
```
sudo ./scripts/landscape-client -c root-client.conf
<ctrl-c>
ls -la /tmp/landscape-root/sockets/
<empty listing>
```


Roadmaster:
LGTM!

PR #89, opened 2020-11-11
Scan /etc/apt/trusted.gpg.d when verifying release upgrade.

simpoir:
This fixes a regression of bionic to focal upgrade. (lp:1903776)

Testing instructions:

* sudo apt build-dep . && dpkg-buildpackage -b
* install resulting packages inside a bionic container.
* register against landscape.c.c
* under Computers -> Packages, trigger a release upgrade (last link)
* Wait for activity success. Check `/var/log/landscape/release-upgrader.log`

### DIFF
--- a/landscape/lib/gpg.py
+++ b/landscape/lib/gpg.py
@@ -1,5 +1,4 @@
 import itertools
-import os.path
 import shutil
 import tempfile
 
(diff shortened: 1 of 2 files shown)


maxiberta (approved):
Looks great, thanks. +1

PR #93, opened 2021-05-06
Bump changelog to upstream v21.05 

simpoir:
Bump changelog expecting to push an upstream release to impish with all the current bugfixes.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,50 @@
+landscape-client (21.05-0ubuntu1) impish; urgency=medium
+
+  * New upstream release 19.05:
+    - Fix apt 2.1.5 regression trying to report broken errors.
+    - Replace deprecated tostring() call with tobytes(). (LP: #1911050)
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,50 @@
+landscape-client (21.05-0ubuntu1) impish; urgency=medium
+
+  * New upstream release 19.05:
+    - Fix apt2.1.5 regression trying to report broken errors.
```
maxiberta, line 4: s/apt2.1.5/apt 2.1.5/ ?

maxiberta (approved):
LGTM! +1

paride:
Hi! Just a couple of drive-by comments:
 - The latest d/changelog entry this PR adds says `New upstream release 19.05` but it should be 21.05 I think.
 - Should the new upstream release be tagged before the packaging is updated? Otherwise d/changelog will refer to a "new upstream version" that isn't defined in a clear way. There's no `21.05` tag at the moment, at least not pushed to GitHub.

PR #94, opened 2021-09-11
Translate legacy py27 messages to py3 messages in broker server.

simpoir:
This should address issues with upgrades never reporting back
when migrating to python3 while still having py27 processes. (LP: #1943291)

Testing instructions:

The quick and easy way to verify this is to
- apply [this patch](https://pastebin.ubuntu.com/p/YTbYjFrvNn/)
- edit `root-client.conf` to point to your preferred server
- then call `sudo ./scripts/landscape-client -c root-client.conf`
- then trigger a release upgrade (in the UI computer packages page, bottom-most link "release upgrade")
- and wait a few minutes for the NOOP update to succeed

The alternative would be to build the client package, create a local repo, and uncomment the extra source during the upgrade.

### DIFF
--- a/landscape/client/broker/server.py
+++ b/landscape/client/broker/server.py
@@ -46,6 +46,7 @@ Diagram::
 import logging
 
 from twisted.internet.defer import Deferred
+from twisted.python.compat import _PY3
 
 from landscape.lib.twisted_util import gather_results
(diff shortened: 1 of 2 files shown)


maxiberta (approved):
LGTM +1

PR #97, opened 2021-11-18
Ignore virtual interfaces in sysinfo Launchpad #1944204

silverdrake11:


maxiberta (commented):
Please, correct me if I'm wrong; are aliases supposed to show up by default now? A quick test:
```bash
$ sudo ifconfig wlan0:1 192.168.100.100
$ ls /sys/devices/virtual/net/wlan*
zsh: no matches found: /sys/devices/virtual/net/wlan*
$ python -c "import netifaces; print(netifaces.interfaces())"
['lo', 'wlan0', 'enx083a88615013', 'lxdbr0', 'veth7af8c96a', 'veth2cebdd17', 'wlan0:1']
```
Changes otherwise make sense and look great.

silverdrake11:
Thanks @maxiberta I copied the alias code back in

maxiberta (approved):
LGTM, thanks! +1

landscape/lib/network.py
```diff
@@ -138,20 +139,30 @@ def get_flags(sock, interface):
...
+        interfaces = os.listdir('/sys/devices/virtual/net')
+    except OSError:
+        interfaces = []
+    return interfaces
+
+
+def get_active_device_info(skip_alias=True, extended=False):
     """
     Returns a dictionary containing information on each active network
     interface present on a machine.
```
simpoir, line 157: This comment isn't accurate anymore, since you now list "active **physical** network interfaces".

landscape/lib/network.py
```diff
@@ -138,20 +139,30 @@ def get_flags(sock, interface):
...
     results = []
+    virtual_interfaces = set(get_virtual())
     try:
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM,
                              socket.IPPROTO_IP)
         for interface, ifaddresses in get_active_interfaces():
-            if interface in skipped_interfaces:
-                continue
-            if skip_vlan and "." in interface:
+            if interface in virtual_interfaces:
```
simpoir, line 165: Have you looked at the various callsites of this function?
It looks like this is used by a networkdevice monitor, which would alter monitoring of those interfaces by landscape-client.
This looks like it's altering a bit more than just sysinfo.

landscape/lib/network.py
```diff
@@ -178,7 +189,7 @@ def get_active_device_info(skipped_interfaces=("lo",),
             if netifaces.AF_INET in ip_addresses or extended:
                 results.append(interface_info)
     finally:
-        del sock
+        sock.close()
```
simpoir, line 192: Might as well drop that try block and use contextmanagers:
```
with socket.socket():
   ...
```

simpoir (changes requested):

simpoir:
There are a few things which bug me with this change, but the most important is the side effects to client monitoring, which is not related to sysinfo.

Also, I don't think the filtering of virtual interfaces is quite the right approach. In some network configurations, the physical interfaces have no addresses and the "main" IP is actually on a vlan or bridge. (if you have a DDWRT home router, that's what you'll see). The suggestion from the bug report of checking the default route interface almost always point to the interface of interest (unless the machine is air-gapped).

silverdrake11:
It seemed simpler to me to do it this way reading the virtual devices, but it sounds like using the default routes is the way to go.

For the side effects, I could add a flag so that `monitor/swiftusage.py` and `monitor/networkdevice.py` is left unchanged. However before my change client monitoring was already filtering out `lo`, vlan interfaces (any that contain `.`) and alias interfaces. Should these continue to be filtered as it was before?

landscape/lib/network.py
silverdrake11: I'm just gonna leave this alone as there seem to be some doubts whether the context manager is supported for sockets in versions of Python earlier than 3.5 https://stackoverflow.com/questions/16772465/how-to-use-socket-in-python-as-a-context-manager

landscape/lib/tests/test_network.py
```diff
@@ -148,6 +148,25 @@ def test_ipv6_only_in_extended_mode(
...
+        default_iface = "test_iface"
+        mock_get_default_interfaces.return_value = [default_iface]
+        mock_get_network_interface_speed.return_value = (100, True)
+        mock_get_flags.return_value = 4163
+        mock_interfaces.return_value = [default_iface]
+        mock_ifaddresses.return_value = {AF_INET: [{"addr": "192.168.0.50"}]}
+        device_info = get_active_device_info(extended=False, default_only=True)
+        interfaces = [i["interface"] for i in device_info]
+        self.assertIn(default_iface, interfaces)
+        self.assertEqual(len(interfaces), 1)
```
simpoir, line 168: nitpick: since you properly isolated this test, the content `interfaces` should be fixed and could be a single assertEqual() on a set.

simpoir (approved):
+1 feel free to disregard the nitpick.

Also, to answer your question, the "default" special gateway value is always there, as per docs.
Works fine with multi ipv6 v4 and other weird setups I could throw at it (dual routes, metrics, bridging).
I think there are still odd cases (or perhaps netiface is smarter than my fake network loops), but for the purpose of sysinfo this new behaviour is desirable.

PR #98, opened 2021-12-02
Fix slow sysinfo when large number of interfaces present #1950311

silverdrake11:
From here https://bugs.launchpad.net/landscape-client/+bug/1950311

### DIFF
--- a/landscape/lib/network.py
+++ b/landscape/lib/network.py
@@ -35,32 +35,16 @@ def is_up(flags):
     return flags & 1
 
 
-def get_active_interfaces():
-    """Generator yields (active network interface name, address data) tuples.
+def is_active(ifaddresses):
+    """Checks if interface address data has an IP address
 
-    Address data is formatted exactly like L{netifaces.ifaddresses}, e.g.::
-
-        ('eth0', {
-            AF_LINK: [
-                {'addr': '...', 'broadcast': '...'}, ],
-            AF_INET: [
-                {'addr': '...', 'broadcast': '...', 'netmask': '...'},
-                {'addr': '...', 'broadcast': '...', 'netmask': '...'},
-                ...],
-            AF_INET6: [
-                {'addr': '...', 'netmask': '...'},
-                {'addr': '...', 'netmask': '...'},
-                ...], })
-
-    Interfaces with no IP address are ignored.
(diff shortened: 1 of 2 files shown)


landscape/lib/network.py
```diff
@@ -35,32 +35,16 @@ def is_up(flags):
...
-    for interface in netifaces.interfaces():
-        ifaddresses = netifaces.ifaddresses(interface)
-        # Skip interfaces with no IPv4 or IPv6 addresses.
-        inet_addr = ifaddresses.get(netifaces.AF_INET, [{}])[0].get('addr')
-        inet6_addr = ifaddresses.get(netifaces.AF_INET6, [{}])[0].get('addr')
-        if inet_addr or inet6_addr:
-            yield interface, ifaddresses
+    inet_addr = ifaddresses.get(netifaces.AF_INET, [{}])[0].get('addr')
+    inet6_addr = ifaddresses.get(netifaces.AF_INET6, [{}])[0].get('addr')
+    if inet_addr or inet6_addr:
```
simpoir, line 45: needless branching…
`return inet_addr or inet6_addr`

landscape/lib/tests/test_network.py
```diff
@@ -173,25 +173,32 @@ def test_skip_loopback(self):
...
+        mock_interfaces.side_effect = lambda: (
+            _interfaces() + ["eth0:foo"])
         device_info = get_active_device_info()
         interfaces = [i["interface"] for i in device_info]
         self.assertNotIn("eth0:foo", interfaces)
 
+    @patch("landscape.lib.network.netifaces.ifaddresses")
+    @patch("landscape.lib.network.netifaces.interfaces")
+    def test_no_extra_netifaces_calls(self, mock_interfaces, mock_ifaddresses):
+        """Make sure filtered out interfaces aren't used in netiface calls"""
```
simpoir, line 197: "…due to their impact on sysinfo/login time with a large amount of interfaces."

It not obvious otherwise that this test is gating performance regression.

simpoir (approved):
+1 with minor changes.

landscape/lib/network.py
silverdrake11: I feel like I should leave this in due to the explicit return value, otherwise it would be a string or None instead of True/False

PR #100, opened 2022-01-04
Computer tags message plugin

silverdrake11:
Launchpad is here https://bugs.launchpad.net/landscape-client/+bug/1940786

### DIFF
--- a/landscape/client/monitor/config.py
+++ b/landscape/client/monitor/config.py
@@ -6,7 +6,7 @@ ALL_PLUGINS = ["ActiveProcessInfo", "ComputerInfo",
                "Temperature", "PackageMonitor", "UserMonitor",
                "RebootRequired", "AptPreferences", "NetworkActivity",
                "NetworkDevice", "UpdateManager", "CPUUsage", "SwiftUsage",
-               "CephUsage"]
+               "CephUsage", "ComputerTags"]
 
 
 class MonitorConfiguration(Configuration):
--- a/landscape/client/monitor/tests/test_computertags.py
+++ b/landscape/client/monitor/tests/test_computertags.py
@@ -1,4 +1,4 @@
-import sys
+import mock
 
 from landscape.client.monitor.computertags import ComputerTags
 from landscape.client.tests.helpers import MonitorHelper, LandscapeTest
@@ -30,12 +30,9 @@ class ComputerTagsTest(LandscapeTest):
         tags = 'check,linode,profile-test'
         file_text = "[client]\ntags = {}".format(tags)
         filename = self.makeFile(file_text)
-        saved_argv = sys.argv
-        try:
-            sys.argv = ['hello.py', '--config', filename]
+        testargs = ["hello.py", "--config", filename]
+        with mock.patch.object(ComputerTags, '_argv', testargs):
             data = self.plugin.get_data()
-        finally:
-            sys.argv = saved_argv
         self.assertEqual(data, tags)
 
     def test_tags_message_sent(self):
--- a/landscape/client/monitor/tests/test_computertags.py
+++ b/landscape/client/monitor/tests/test_computertags.py
@@ -1,4 +1,3 @@
-import mock
 import sys
 
 from landscape.client.monitor.computertags import ComputerTags
@@ -31,9 +30,12 @@ class ComputerTagsTest(LandscapeTest):
         tags = 'check,linode,profile-test'
         file_text = "[client]\ntags = {}".format(tags)
         filename = self.makeFile(file_text)
-        testargs = ["hello.py", "--config", filename]
-        with mock.patch.object(sys, 'argv', testargs):
+        saved_argv = sys.argv
+        try:
+            sys.argv = ['hello.py', '--config', filename]
             data = self.plugin.get_data()
+        finally:
+            sys.argv = saved_argv
         self.assertEqual(data, tags)
 
     def test_tags_message_sent(self):
--- a/landscape/client/monitor/tests/test_computertags.py
+++ b/landscape/client/monitor/tests/test_computertags.py
@@ -1,3 +1,6 @@
+import mock
+import sys
+
 from landscape.client.monitor.computertags import ComputerTags
 from landscape.client.tests.helpers import MonitorHelper, LandscapeTest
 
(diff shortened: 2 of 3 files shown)


silverdrake11:
What should happen if tags are invalid? During registration if this happens, then None is sent and an error is logged. Currently I am doing the same thing but maybe the invalid tag should be filtered out instead?

landscape/client/monitor/config.py
```diff
@@ -6,7 +6,7 @@
                "Temperature", "PackageMonitor", "UserMonitor",
                "RebootRequired", "AptPreferences", "NetworkActivity",
                "NetworkDevice", "UpdateManager", "CPUUsage", "SwiftUsage",
-               "CephUsage"]
+               "CephUsage", "ComputerTags"]
```
simpoir, line 9: I'm actually on the fence over this. Having things sync automatically for new clients would be nice, but if you add it to `ALL_PLUGINS`, then it becomes enabled by default for any existing client, whether you want it or not. The only way to not have it appear in your workflow would be to set `monitor_plugins=all,the,other,plugins`, which is awkward.

That's probably fine for almost all cases use cases, as client-provided tags are not used a lot in the wild. Clients suddenly messing with my tags after an upgrade could be considered a regression.

Any thoughts on this?

landscape/client/monitor/tests/test_computertags.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        If invalid tag detected then message contents should be None
+        """
+        tags = 'check,lin ode'
+        file_text = "[client]\ntags = {}".format(tags)
+        config_filename = self.config.default_config_filenames[0]
+        self.makeFile(file_text, path=config_filename)
+        self.assertEqual(self.plugin.get_data(), None)
+
+    def test_empty_config_file(self):
```
simpoir, line 48: Not necessarily empty config, but empty tags sure.

Client wouldn't actually start with an empty config.

landscape/client/monitor/tests/test_computertags.py
```diff
@@ -0,0 +1,52 @@
...
+    helpers = [MonitorHelper]
+
+    def setUp(self):
+        super(ComputerTagsTest, self).setUp()
+        self.plugin = ComputerTags()
+        self.monitor.add(self.plugin)
+
+    def test_tags_are_read(self):
+        """
+        Tags are read from the default config file
```
simpoir, line 16: tags should be read from the currently used config file, not just the default one.

simpoir (changes requested):
see inline comments.

landscape/client/monitor/config.py
silverdrake11: I think it should be fine because the clients suddenly messing would just be adding tags if you changed the config file. Nothing would get deleted.

landscape/client/monitor/tests/test_computertags.py
silverdrake11: I added a test case to cover this

simpoir (approved):
+1

Could have done without mocking, since it's a class attribute which can be just set on test instances without having lasting side-effects. But it's fine.

landscape/client/monitor/tests/test_computertags.py
```diff
@@ -13,6 +10,7 @@ def setUp(self):
         super(ComputerTagsTest, self).setUp()
         self.plugin = ComputerTags()
         self.monitor.add(self.plugin)
+        ComputerTags._argv = ['hello.py']  # For testing sys.argv
```
simpoir, line 13: My point was more to use shadowing on the instance: 
`self.plugin._argv = ["hello.py"]`
would override/shadow just that instance, without any subsiding side-effect.

Having attributes on classes and shadowing the default on test instances is a common pattern.

simpoir: Something about leveraging the python MRO which looks-up `object -> class -> superclass`

silverdrake11: Right, changing it to shadow the instance did break get_data() because it was using the class attribute not the instance one. To avoid these issues I used your original idea about default arguments in the __init__

simpoir (approved):
still +1

asangeethada:
I am having the same issue still, ComputerTags plugin does not seem to be loading when I look at /vat/log/landscape/monitor.log 


`Monitor started with config /etc/landscape/client.config
2023-06-07 13:28:23,167 INFO     [MainThread] Registering plugin landscape.client.monitor.activeprocessinfo.ActiveProcessInfo.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.computerinfo.ComputerInfo.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.loadaverage.LoadAverage.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.memoryinfo.MemoryInfo.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.mountinfo.MountInfo.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.processorinfo.ProcessorInfo.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.temperature.Temperature.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.packagemonitor.PackageMonitor.
2023-06-07 13:28:23,168 INFO     [MainThread] Registering plugin landscape.client.monitor.usermonitor.UserMonitor.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.rebootrequired.RebootRequired.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.aptpreferences.AptPreferences.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.networkactivity.NetworkActivity.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.networkdevice.NetworkDevice.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.updatemanager.UpdateManager.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.cpuusage.CPUUsage.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.swiftusage.SwiftUsage.
2023-06-07 13:28:23,169 INFO     [MainThread] Registering plugin landscape.client.monitor.cephusage.CephUsage.
2023-06-07 13:28:28,423 INFO     [MainThread] This machine does not appear to be a Swift machine. Deactivating plugin.
2023-06-07 13:28:53,212 INFO     [MainThread] This machine does not appear to be a Ceph machine. Deactivating plugin.
2023-06-07 13:29:13,134 INFO     [MainThread] Got notification of impending exchange. Notifying all plugins.
2023-06-07 13:30:23,104 INFO     [MainThread] Got notification of impending exchange. Notifying all plugins.`

landscape-client --version
18.01-0ubuntu13

PR #101, opened 2022-01-12
faster Processes sysinfo

collinanderson:
This makes process and zombie counting much much faster, especially when there's a lot of processes.

This code runs on every login via motd.

### DIFF
--- a/landscape/sysinfo/processes.py
+++ b/landscape/sysinfo/processes.py
@@ -1,6 +1,6 @@
-from twisted.internet.defer import succeed
+import os
 
-from landscape.lib.process import ProcessInformation
+from twisted.internet.defer import succeed
 
 
 class Processes(object):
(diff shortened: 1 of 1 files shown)


landscape/sysinfo/processes.py
```diff
@@ -14,10 +14,20 @@ def register(self, sysinfo):
...
+        for pid in os.listdir(self._proc_dir):
+            if not pid.isdigit():
+                continue
+            status_path = os.path.join(self._proc_dir, pid, "status")
+            try:
+                fd = os.open(status_path, os.O_RDONLY)
+                try:
+                    data = os.read(fd, 2048)
+                finally:
+                    os.close(fd)
```
simpoir, line 26: The whole block should use "modern" context manager. Also, the use of human-readable `/status` instead of the machine parse-able `/stat` seems like a poor choice. I assume this was just copied and trimmed from l/lib/process.py, but the proper way would probably be like:

```
try:
    with open(status_path, "rb") as fd:
        data = f.read()
except IOError:
    continue
num_processes += 1
if b"Z" == data.split(b" ", 3)[2]:
    num_zombies += 1
```

One or two zombies tests in `landscape/sysinfo/tests/test_processes.py` would need an extra `create_data(…, stat_data="0 0 Z 0 0 0 0")` argument to make them pass.

simpoir (dismissed):
The tab separator proposed is actually incorrect, and it makes tests fail.

juanmitaboada:
I have remade this PR by applying suggestions from @simpoir at #155

srunde3:
I'm going to close this PR as it was superseded by (now merged) PR 155: https://github.com/canonical/landscape-client/pull/155

PR #106, opened 2022-03-03
Fixes import issues with jammy. (LP: #1962539)

simpoir:
* replace deprecated twisted _PY3
* replace legacy types which are gone (unicode, long)

Testing instructions:

For the bug itself, running `./scripts/landscape-sysinfo` from within a
jammy container should be sufficient.
Since this branch touches practically every code path, I would recommend
fully launching the client and registering it.


silverdrake11 (approved):
LGTM +1

PR #107, opened 2022-03-10
Broker back off when receiving 5xx responses LP: #1947399

silverdrake11:
The original bug is here https://bugs.launchpad.net/landscape-client/+bug/1947399

### DIFF
--- a/landscape/client/broker/exchange.py
+++ b/landscape/client/broker/exchange.py
@@ -591,8 +591,16 @@ class MessageExchange(object):
 
             if isinstance(error, HTTPCodeError):
                 if error.http_code == 429 or (500 <= error.http_code <= 599):
-                    # We add an exponentially increasing delay if the server
-                    # is overloaded to decrease load
+                    # We add an exponentially increasing delay ("backoff") if
+                    # the server is overloaded to decrease load. We assume that
+                    # any server backend error is deserving backoff, including
+                    # 429 which is sent from the server on purpose to trigger
+                    # the backoff. Whether the server is down or overloaded
+                    # (503), has a server bug crashing the response (500),
+                    # backing off should have no ill effect and help the
+                    # service to recover. Client-configuration related errors
+                    # (501, 505) are probably fine to throttle too as a higher
+                    # rate won't help resolve them
                     self._backoff_counter.increase()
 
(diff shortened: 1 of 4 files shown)


silverdrake11:
Manual testing instructions:
- Modify or mock server to throw 500s
- Set start delay to something smaller than 5 min in client
- Watch messages

landscape/client/broker/exchange.py
```diff
@@ -642,6 +651,7 @@ def schedule_exchange(self, urgent=False, force=False):
                 interval = self._config.urgent_exchange_interval
             else:
                 interval = self._config.exchange_interval
+            interval += self._backoff_counter.get_random_delay()
```
simpoir, line 654: shouldn't you pass self._config.stagger here to use the same random ratio as configured, instead of the arbitrary ratio you have?

simpoir (approved):
+1
I'm not blocking this branch, but see the few inconsistencies in the comments.

landscape/lib/backoff.py
```diff
@@ -0,0 +1,53 @@
...
+                4      40
+                5      60 (max)
+        '''
+        if self._error_count:
+            delay = (2 ** (self._error_count - 1)) * self._start_delay
+        else:
+            delay = 0
+        return min(int(delay), self._max_delay)
+
+    def get_random_delay(self, percent_rand=0.25):
```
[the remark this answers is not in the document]
silverdrake11: Fixed

landscape/client/broker/exchange.py
silverdrake11: In this specific case, although 25% is an arbitrary random stagger, I don't think it's a good idea of the user to control this value, as it's more of an implementation detail, which its only purpose is so that requests don't hit the server at the same time if it is under heavy load. However you bring up a good point, since none of this is not currently configurable, maybe we should let the the start and stop be configurable?

simpoir: The config default value is pretty arbitrary too. However, I think you're right: There is no need to have this as user-configurable since it shouldn't stagger under normal circumstances and, unlike the stagger configuration, you can just bypass any delay by restarting the client.

Ignore my previous comment.

PR #108, opened 2022-03-22
Don't show updates available in motd for registered landscape-managed…

Perfect5th:
… machines (LP: #1621978)

### DIFF
--- a/debian/landscape-client.postinst
+++ b/debian/landscape-client.postinst
@@ -139,6 +139,12 @@ END
             fi
             rm -f $DEFAULTS
         fi
+
+        # Since motd shows a different number of security updates,
+        # suppress the message if registered.
+        if landscape-config --is-registered >/dev/null 2>&1; then
+            dpkg-divert --package $PACKAGE --rename /etc/update-motd.d/90-updates-available
+        fi
     ;;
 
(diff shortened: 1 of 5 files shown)


debian/landscape-client.postinst
```diff
@@ -139,6 +139,12 @@ END
             fi
             rm -f $DEFAULTS
         fi
+
+        # Since motd shows a different number of security updates,
+        # suppress the message if registered.
+        if landscape-config --is-registered >/dev/null 2>&1; then
```
simpoir, line 145: Nice try, but that won't work. The chances the client is already registered before being installed are pretty slim.

Perfect5th: Fair enough - I guess a second install would have to be run, which isn't going to happen any time soon.

Perfect5th:
Manual testing steps:

- build landscape-common and landscape-client debs and install
- run `sudo run-parts /etc/update-motd.d/`, info about updated packages is shown
- register client with landscape server
- run `sudo run-parts /etc/update-motd.d/` again, info about updated packages is not shown
- uninstall landscape-client
- run `sudo run-parts /etc/update-motd.d/` again, info about updated packages is shown again

PR #110, opened 2022-04-07
Use systemctl to signal post-logrotate.

simpoir:
Use systemctl to signal parent process instead of using pidfile (LP: #1968189)
Only signal once for all rotations.

Testing instructions:

```
apt-get build-dep .
dpkg-buildpackage -b
dpkg -i ../landscape*.deb
systemctl start landscape-client  # it's unconfigured, but this is fine.
tail -F /var/log/landscape/*.log &
logrotate --force /etc/logrotate.conf  
```
You'll get a  couple of outputs, one per landscape process log, telling the log files were rotated.

### DIFF
--- a/debian/landscape-client.logrotate
+++ b/debian/landscape-client.logrotate
@@ -5,8 +5,9 @@
     notifempty
     compress
     nocreate
+    sharedscripts
     postrotate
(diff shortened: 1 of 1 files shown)


Perfect5th (approved):
tested locally and looks good to me.

Perfect5th (approved):

PR #111, opened 2022-06-15
Disable stopping service during upgrade.

simpoir:
The client package using no-start (due to being enabled from
landscape-config), avoid stopping the service on upgrade so that

1. landscape may upgrade itself and see the upgrade complete
2. the client doesn't stay down after upgrade (due to no-start, dh will never restart it)


Perfect5th (approved):
lgtm

PR #115, opened 2022-10-05
Replace dependency on /etc/lsb-release with /usr/bin/lsb_release to permit Debian compatibility - #1990268

Borgquite:
As requested in Bug #1990268 here is some code that should replace the /etc/lsb-release dependency (which is Ubuntu-specific) with more generic /usr/bin/lsb_release dependency, allowing package reporting to work on Debian.

Looking through the code there are various references in the tests - I'm afraid this is beyond my skills, but hoping someone with more knowledge might be able to help out here!

### DIFF
--- a/landscape/lib/lsb_release.py
+++ b/landscape/lib/lsb_release.py
@@ -1,19 +1,58 @@
 """Get information from /usr/bin/lsb_release."""
+import os
+from subprocess import DEVNULL, CalledProcessError, check_output
 
-LSB_RELEASE_FILENAME = "/usr/bin/lsb_release"
-LSB_RELEASE_INFO_KEYS = {"distributor-id": "-si",
-                         "release": "-sr",
-                         "code-name": "-sc",
-                         "description": "-sd"}
+LSB_RELEASE = "/usr/bin/lsb_release"
+LSB_RELEASE_FILENAME = "/etc/lsb_release"
+LSB_RELEASE_FILE_KEYS = {
+    "DISTRIB_ID": "distributor-id",
+    "DISTRIB_DESCRIPTION": "description",
+    "DISTRIB_RELEASE": "release",
+    "DISTRIB_CODENAME": "code-name",
+}
 
-import subprocess
 
-def parse_lsb_release(lsb_release_filename):
-    """Return a C{dict} holding information about the system LSB release.
-    @raises: An OSError exception if C{lsb_release_filename} does not exist.
+def parse_lsb_release(lsb_release_filename=None):
+    """
+    Returns a C{dict} holding information about the system LSB release.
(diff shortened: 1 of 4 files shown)


Perfect5th:
The tests need to pass for the PR to pass our CI checks.

If you've allowed edits by maintainers on this PR, I'll fix the tests.

Perfect5th:
@Borgquite I'd appreciate it if you'd to test my latest changes on debian.

landscape/lib/lsb_release.py
```diff
@@ -1,26 +1,58 @@
...
+    Returns a C{dict} holding information about the system LSB release.
+    Reads from C{lsb_release_filename} if it exists, else calls
+    C{LSB_RELEASE}
+    """
+    if lsb_release_filename and os.path.exists(lsb_release_filename):
+        return parse_lsb_release_file(lsb_release_filename)
+
+    try:
+        lsb_info = check_output([LSB_RELEASE, "-as"], stderr=DEVNULL)
+    except CalledProcessError:
```
silverdrake11, line 26: I tried this and I am getting FileNotFound error instead of CalledProcessError if lsb-release is not installed on the system so it doesn't fall back to reading the file

Perfect5th: Right. overlooked that, will fix.

silverdrake11: I think I would prefer to using one method, probably the command line. It keeps it simple to use one method and also in debian falling back to the file is not going to work, (unless you use the /etc/os-release but then it's gonna be different keys for it). lsb-release should be installed in ubuntu and I'm fine with the release information missing from server if this is not installed, but I can also understand having it be foolproof in ubuntu  so your call

silverdrake11 (changes requested):

Perfect5th:
I'm going to respectfully decline restricting it to one method, but not only for "foolproofness". Currently a lot of the application code passes in `lsb_release_filename`, so not having it handle the file-reading case necessitates a _lot_ of additional touches to the code-base.

landscape/lib/lsb_release.py
Perfect5th: I'm going to respectfully decline restricting it to one method, but not only for "foolproofness". Currently a lot of the application code passes in `lsb_release_filename`, so not having it handle the file-reading case necessitates a _lot_ of additional touches to the code-base.

silverdrake11 (approved):

Borgquite:
@Perfect5th Tested on Debian 10. It didn't work because of the dependence on DEVNULL, which is only in Python > 3.3 (Debian 10 uses Python 2.7.16
Reverted to os.devnull, seems to work OK now?

Borgquite:
Great! Will look forward to this being released :)

PR #118, opened 2022-10-07
Add Ubuntu Pro Info monitor

Perfect5th:


silverdrake11:
This looks good. The only thing I noticed is the run_interval is missing. It's using the default I think which is too frequent. Should be like once every 60 or 15 minutes.

Perfect5th:
I think 15 minutes makes sense. It's pretty CPU-light, so more frequent than 60 minutes is ok.

As for that bug... I'm not overly concerned. I doubt it'll happen often, and re-registering isn't the end of the world. As you said, it won't matter once the full pro changes are in place.

landscape/client/monitor/ubuntuproinfo.py
```diff
@@ -0,0 +1,47 @@
...
+    Plugin that captures and reports Ubuntu Pro registration
+    information.
+
+    We use the `ua` CLI with output formatted as JSON. This is sent
+    as-is and parsed by Landscape Server because the JSON content is
+    considered "Experimental" and we don't want to have to change in
+    both Client and Server in the event that the format changes.
+    """
+
+    run_interval = 15
```
silverdrake11, line 18: Should be 900 seconds

silverdrake11 (changes requested):

landscape/client/monitor/ubuntuproinfo.py
Perfect5th: oops - fixed.

silverdrake11 (approved):
LGTM +1

PR #119, opened 2022-10-12
filter out tap network interfaces in sysinfo

Perfect5th:
Also slight restructuring of the methods in a way that I hope makes adding future filters more composable.


landscape/lib/network.py
```diff
@@ -124,58 +124,93 @@ def get_default_interfaces():
...
                 continue
-            interface_info = {"interface": interface}
-            interface_info["flags"] = flags
-            speed, duplex = get_network_interface_speed(
-                sock, interface.encode())
-            interface_info["speed"] = speed
-            interface_info["duplex"] = duplex
+
             ip_addresses = get_ip_addresses(ifaddresses)
+            if not (netifaces.AF_INET in ip_addresses or extended):
```
silverdrake11, line 155: Could be if not extended:
                    if AF_INET not in ip_adrresses:
                          continue

landscape/lib/network.py
```diff
@@ -124,58 +124,93 @@ def get_default_interfaces():
...
+    def filter_vlan(interface):
+        return "." in interface
+
+    def filter_alias(interface):
+        return ":" in interface
+
+    default_ifs = get_default_interfaces()
+
+    def filter_default(interface):
+        return default_only and interface not in default_ifs
```
silverdrake11, line 198: It might not be clear to others why get_default_interfaces() is defined outside of function. Maybe add a comment there that we do not want to call this more than once

landscape/lib/tests/.test_network.py.~undo-tree~
```diff
@@ -0,0 +1,181 @@
+(undo-tree-save-format-version . 1)
```
silverdrake11, line 1: Forgot to exclude this file

silverdrake11 (changes requested):

landscape/lib/tests/.test_network.py.~undo-tree~
Perfect5th: oops

landscape/lib/network.py
Perfect5th: de morganing it to `if not extended and netifaces.AF_INET not in ip_addresses` seems like a good compromise.

landscape/lib/network.py
Perfect5th: Added comment.

silverdrake11 (approved):
LGTM +1

PR #120, opened 2022-10-16
Round load average to 2 decimal places

prpr19xx:
More recent kernels (e.g. 5.15 in Ubuntu 22) seem to give many more than the 2 digits of precision that older ones (e.g. 5.4 in Ubuntu 20) gave in the return from getloadavg().
This restores the backwards compatibility and prevents silly things like this from landscape-sysinfo:
System load: 0.080078125   Processes: 124

### DIFF
--- a/landscape/sysinfo/load.py
+++ b/landscape/sysinfo/load.py
@@ -9,5 +9,6 @@ class Load(object):
         self._sysinfo = sysinfo
 
     def run(self):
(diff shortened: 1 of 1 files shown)


Perfect5th (approved):
LGTM apart from lint.

prpr19xx:
Thanks. Didn't know about the line length limit at 79.

Perfect5th:
Actually not knowing about it is a good point. I'm going to add a mention of `make lint` to the README.

PR #121, opened 2022-10-17
Mention check3 and linting in README for contributors

Perfect5th:
We get a few PRs that fail the PR checks (usually on lint). We can lessen how many of these we get by at least _mentioning_ that contributors can run `check3` and `lint` locally before they create a PR.


silverdrake11 (approved):
LGTM +1

PR #122, opened 2022-10-19
Remove deprecated apt-key to stop gpg warning (LP: #1973202)

silverdrake11:
Manual testing instructions:

1) Register client to on prem server
2) Follow steps to set up mirror (use restricted component to limit file size) all the way to "Associate computers with repository profile" making sure to associate the profile with the registered client https://docs.ubuntu.com/landscape/en/repositories
3) Look in /etc/apt/trusted.gpg.d for the keys and confirm no 'trusted.gpg' file gets created


Perfect5th (approved):
LGTM

PR #125, opened 2022-11-17
Separate bytes-handling for python2 and python3

Perfect5th:
This patch separates the coercion of bytes-type schema data in py2 from py3. The py2 behaviour keeps the previous behaviour of raising an exception if the data is not bytes, but the py3 behavious will encode the data as bytes if it is a py3 unicode str.

### DIFF
--- a/landscape/lib/schema.py
+++ b/landscape/lib/schema.py
@@ -1,4 +1,5 @@
 """A schema system. Yes. Another one!"""
+import six
 from twisted.python.compat import iteritems, unicode, long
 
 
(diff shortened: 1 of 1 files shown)


landscape/lib/schema.py
```diff
@@ -1,4 +1,5 @@
 """A schema system. Yes. Another one!"""
+import six
```
silverdrake11, line 2: Seems like six is a dependency of landscape server. I don't see used anywhere in client. I'm not totally sure if six is pre-installed in all places where we can run landscape cilent

silverdrake11:
LGTM +1

silverdrake11 (approved):

PR #126, opened 2022-11-18
Trimmed big message error and result text (LP: #1994534)

silverdrake11:
This is a fix for the client side

Manual testing instructions:
1) Send a large message to server by modifying landscape/client/package/reporter.py and force_apt_update config (same as required  for testing server changes)
1) Confirm message was truncated by looking in package reporter errors in the alerts in server

### DIFF
--- a/landscape/client/broker/tests/test_exchange.py
+++ b/landscape/client/broker/tests/test_exchange.py
@@ -151,8 +151,8 @@ class MessageExchangeTest(LandscapeTest):
         """
         self.mstore.set_accepted_types(["package-reporter-result"])
         self.exchanger._max_log_text_bytes = 5
-        self.exchanger.send({"type": "package-reporter-result", "err":"E"*10, 
-                             "code":0})
+        self.exchanger.send({"type": "package-reporter-result", "err": "E"*10,
+                             "code": 0})
         self.exchanger.exchange()
         self.assertEqual(len(self.transport.payloads), 1)
         messages = self.transport.payloads[0]["messages"]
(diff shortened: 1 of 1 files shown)


silverdrake11:
Client bug report is here https://bugs.launchpad.net/landscape-client/+bug/1995775 (server side one is in title)

landscape/client/broker/exchange.py
```diff
@@ -438,6 +439,20 @@ def _message_is_obsolete(self, message):
 
         return result
 
+    def truncate_message_field(self, field, message):
+        """
+        Truncates message field value based on length
+        """
+        if field in message:
+            value = message[field]
+            if isinstance(value, str):
```
Perfect5th, line 448: Is there a reason we are only truncating string fields? What about `bytes`? I wonder if it might be more valuable to add a `max_length` init parameter (with a default of `None`) to these schema classes https://github.com/CanonicalLtd/landscape-client/blob/master/landscape/lib/schema.py#L67-L102
and do the truncation on coercion.

silverdrake11: Those two fields being truncated are defined as Unicode() in the schema, so I added the extra check there to prevent trying to truncate a None value. 

If it's in the coerce, then I'm worried about truncating a json or something that is parsed. Those two fields I chose seem to only be for informational purposes

Perfect5th: Ok. If both fields are `Unicode`, and since we no longer support Python 2. All good.

Perfect5th (approved):

PR #130, opened 2022-12-13
Added livepatch status to client

silverdrake11:


landscape/client/monitor/livepatch.py
```diff
@@ -0,0 +1,71 @@
...
+    run_interval = 1800  # Every 30 min
+
+    def get_data(self):
+        json_output = get_livepatch_status('json')
+        readable_output = get_livepatch_status('humane')
+        return json.dumps({'humane': readable_output, 'json': json_output},
+                          sort_keys=True)  # Prevent randomness for cache
+
+
+def get_livepatch_status(format_type):
```
Perfect5th, line 28: I think it would be cleaner if you had two methods, `get_livepatch_status_json` and `get_livepatch_status_humane`. Sure it would duplicate the `subprocess.run` call, but it would de-nest this method and untangle the parsing logic/cleanup for each that you're currently doing in the current method.

Additionally, it sets a better precedent if we add other formats (like, actual `yaml`). However, needing another machine-readable format seems unlikely.

I'm fine with this as-is - it's just something to think about. If you disagree, no worries.

Perfect5th (approved):

PR #131, opened 2022-12-22
More verbose error

juanmitaboada:
It took me a while to discover that I had a duplicate entry in my configuration. With this small improvement, people will be pointed in the right direction.


Perfect5th (approved):
LGTM

PR #136, opened 2023-02-21
Do not overwrite sources.list.d items when handling sources changes.

Perfect5th:


silverdrake11:
What do you think about a True or False flag in the landscape client config, that would control this behavior?

Perfect5th:
> 

Yes, I like that. I think it should default to the "previous" behaviour - feels a little like feature-flagging it. Once the behaviour is more nuanced we can do away with the config.

silverdrake11:
LGTM +1

silverdrake11 (approved):

PR #139, opened 2023-02-28
Security monitor

juanmitaboada:
This monitor aims to manage all the security reports from the client, right now it supports:

- Listening ports (ports that are opened and in LISTENING state)
- Report information from rkhunter (checked/suspicious files and checked/suspicious rootkits)


Perfect5th:
This LGTM, and I like it in general. We'll need to roadmap how we want to surface this information in the server UI - likely that won't happen for awhile, as we have a UI overhaul in the works.

I'm also not :100: about calling the message type "rkhunter-info". I know rkhunter is pretty widely-regarded/-used but I'd prefer if we keep the message types a agnostic as to the actual tools being used to provide the data. That way if we swap rkhunter for a different tool in the future the message type still makes sense (unlikely, but still). May I suggest `rootkit-scan-info`?

juanmitaboada:
`rootkit-scan-info` LGTM

PR #140, opened 2023-03-06
Broker is able to bootstrap the landscape-client folder (LP: #1868730)

juanmitaboada:
The problem in [BUG 1868730](https://bugs.launchpad.net/landscape-client/+bug/1868730) may be related to the **broker** not starting up. I have added the bootstrap process to the broker service, so it will ensure that **sockets** folder exists before trying to startup.


Perfect5th (approved):
LGTM. Manually tested and I'm not seeing any serious issues. The watchdog starting _after_ the broker seems to be the root cause of this issue, and making the bootstrapping redundant in both services seems like a safe enough fix.

PR #141, opened 2023-03-06
Last apt-key removal (LP: #1990435)

juanmitaboada:
This is related to https://bugs.launchpad.net/landscape-client/+bug/1990435


juanmitaboada:
I believe there is one more change to get done in landscape-client.postrm

PR #143, opened 2023-03-07
Makefile with pre-commit hook support

juanmitaboada:


Perfect5th (approved):
LGTM, though we should probably in future move pip-installed deps to (at the very least) a requirements.txt.

Perfect5th:
Additionally, if we're in a venv, this'll work. If we're not, this is likely to fail due to permission issues. We SHOULD be in a venv, though.

PR #146, opened 2023-03-15
Removed unnecesary select to the database (LP: #1994951)

juanmitaboada:
This solution also resolves the exception produced by:
https://bugs.launchpad.net/landscape-client/+bug/1994951

### DIFF
--- a/landscape/lib/apt/package/store.py
+++ b/landscape/lib/apt/package/store.py
@@ -329,21 +329,23 @@ class PackageStore(HashIdStore):
     @with_cursor
     def add_task(self, cursor, queue, data):
         data = bpickle.dumps(data)
+        now = time.time()
         cursor.execute(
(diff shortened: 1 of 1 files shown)


landscape/lib/apt/package/store.py
```diff
@@ -450,23 +452,12 @@ def remove(self, cursor):
...
 class PackageTask:
-    def __init__(self, db, id):
+    def __init__(self, db, cid, queue, timestamp, data):
         self._db = db
-        self.id = id
-
-        cursor = db.cursor()
-        try:
-            cursor.execute(
-                "SELECT queue, timestamp, data FROM task " "WHERE id=?",
```
Perfect5th, line 460: Ha looks like I missed a "could be single string" previously. Oh well, doesn't matter now.

Perfect5th (approved):
LGTM

PR #147, opened 2023-03-15
1989968 wrong monitor plugins option in example

patriacaelum:
Resolves [Bug #1989968](https://bugs.launchpad.net/landscape-client/+bug/1989968).

- The HardwareInventory plugin was removed from `example.conf`
- When invalid plugins are specified, a warning is emitted and that plugin is skipped. Prior, the application would crash.


Perfect5th (approved):
LGTM

PR #148, opened 2023-03-16
Added snap for landscape-client

bceverly:
This PR contains the initial snapification of the client.  It will register but not do much else.

### DIFF
--- a/README
+++ b/README
@@ -91,7 +91,7 @@ make lint
 
 First, you need to ensure that you have the appropriate tools installed:
 ```
-$ sudo snap install snapcraft
+$ sudo snap install snapcraft --classic
 $ lxd init --auto
 ```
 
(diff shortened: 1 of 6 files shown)


Perfect5th:
Looks pretty good, but could we get maybe some text in the README about building the snap?

bceverly:
Added details to the README on how to build the snap and push it to the store

README
```diff
@@ -86,3 +86,104 @@ Before opening a PR, make sure to run the full testsuite and lint
 make check3
 make lint
 ```
+
+### Building the Landscape Client snap
+
+First, you need to ensure that you have the appropriate tools installed:
+```
+$ sudo snap install snapcraft
```
Perfect5th, line 94: I think we need `--classic`

Perfect5th (approved):
LGTM

PR #152, opened 2023-04-09
Changed the way Dependency data is readen (LP: #1813442)

juanmitaboada:
I got deeper into this problem and found that it was happening because we were using "version.record" which is the source of the failure.

Since we don't want to mess up with the UnicodeDecodeError still happening in the apt package, I  decided to switch the way the dependencies information is readen from the package, and I removed "version.record" from the code so it doesn't try to parse the broken data anymore.

Also, I discovered in the original source code that "PreDepends" was empty in all my tests. Now it has information as I think it should work. I would like somebody to check my solution in search of some collateral problems with the new "PreDepends".

I also renamed a variable "hash" in facade to "hash_skeleton".

### DIFF
--- a/landscape/lib/apt/package/facade.py
+++ b/landscape/lib/apt/package/facade.py
@@ -265,15 +265,15 @@ class AptFacade:
             if not self._is_main_architecture(package):
                 continue
             for version in package.versions:
-                hash = self.get_package_skeleton(
+                skeleton_hash = self.get_package_skeleton(
                     version,
                     with_info=False,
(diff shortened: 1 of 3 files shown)


Perfect5th:
I think this looks pretty good. I can't think of any "collateral problems". If you think there's a whole in the tests I would be very grateful if you could fix that :wink:

juanmitaboada:
I checked again for some possible "collateral problems", and I think there are none. LGTM

PR #155, opened 2023-04-15
Faster Processes SysInfo - These are the corrections suggested by Sim…

juanmitaboada:
These are the corrections suggested by @simpoir to #101 which has been around for longer than a year


Perfect5th (approved):
LGTM, based on previous review and some personal manual testing

PR #156, opened 2023-04-15
Add landscape-sysinfo CLI arg to specify maximum column width

juanmitaboada:
These are the corrections to #84 which has been around for almost three years


Perfect5th (approved):
LGTM

PR #157, opened 2023-04-18
initial SnapMonitor and SnapManager implementations - installation and removal only

Perfect5th:


Perfect5th:
Let me know if you want to manually test and I can provide instructions (requires a private branch of Landscape Server that I've been working on in parallel).

landscape/client/deployment.py
```diff
@@ -158,6 +158,12 @@ def make_parser(self):
             help="Repository profiles manage the files in "
             "/etc/apt/sources.list.d",
         )
+        parser.add_option(
+            "--snap-monitor-interval",
+            default=30 * 60,
```
silverdrake11, line 163: Could be 15 minutes. It's cached

silverdrake11: Actually looking at the output, looks like there are timestamps in there. So best to strip them out so DataWatcher can cache the output and it's not sent every n minutes. Although the KeyDict selects the needed values, the caching should happen before

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,226 @@
...
+            except SnapdHttpException as e:
+                result = e.json["result"]
+                logging.error("Error installing snap '%s': %s", name, result)
+                errors[(name, revision, channel)] = result
+
+        return self._check_statuses(installing, opid, errors)
+
+    def _handle_remove_snaps(self, message):
+        """
+        Removes the snaps indicated in `message` using the snap interface.
```
silverdrake11, line 75: This function is very similar to _install_snaps, so some code could be reused

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,226 @@
...
+        Uses the snap interface to retrieve the status of a snap change.
+        Polls every `interval` seconds a maximum of `attempts` times, as
+        configured in the registry.
+        """
+        counter = 0
+        attempts = getattr(self.registry.config, "snapd_poll_attempts", 5)
+        interval = getattr(self.registry.config, "snapd_poll_interval", 15)
+        snap_http = SnapHttp()
+
+        def get_status():
```
silverdrake11, line 140: Found there's a twisted LoopingCallWithCount so I think  it passes an integer like `get_status(count)` so can get rid of the manual counter increment

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Installs the snaps indicated in `message` using the snap interface.
+        """
+        snaps = message["snaps"]
+        opid = message["operation-id"]
+        snap_http = SnapHttp()
+        errors = {}
+        installing = []
+
+        logging.info("Installing snaps: %s", snaps)
```
silverdrake11, line 40: "{}".format or f-string is preferable

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,226 @@
...
+
+        def get_status():
+            """
+            Looping function that stashes results in loop.result. Exits when a
+            non-incomplete snap change status is found, or upon an error.
+            """
+            nonlocal counter
+            counter += 1
+
+            logging.debug("Polling snapd for status of change %s", change_id)
```
silverdrake11, line 148: Some of these could be in, info(), would make it easier to go through broker.log since, debug is not enabled by default

landscape/client/monitor/snapmonitor.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def register(self, registry):
+        self.config = registry.config
+        self.run_interval = self.config.snap_monitor_interval
+
+        super(SnapMonitor, self).register(registry)
+
+    def get_data(self):
+        try:
+            snaps = self._snap_http.get_snaps()
```
silverdrake11, line 29: See comment about caching

landscape/client/snap/http.py
```diff
@@ -0,0 +1,145 @@
...
+        response_code = curl.getinfo(curl.RESPONSE_CODE)
+        if response_code >= 400:
+            raise SnapdHttpException(buff.getvalue())
+
+    def _post(self, path, body):
+        """
+        Perform a POST request of `path` to the snap REST API, with the
+        JSON-ified `body`
+        """
+        curl, buff = self._setup_curl(path)
```
silverdrake11, line 125: Could some of these steps be in the init and self.curl be used?

landscape/client/snap/http.py
```diff
@@ -0,0 +1,145 @@
...
+                "channel": channel,
+                "classic": classic,
+            }.items()
+            if v is not None
+        }
+
+        return self._post("/snaps/" + name, body)
+
+    def remove_snap(self, name):
+        """Removes a snap by `name`."""
```
silverdrake11, line 91: Comment is not needed

landscape/lib/schema.py
```diff
@@ -171,26 +171,35 @@ class KeyDict:
...
             raise InvalidError(f"{value!r} is not a dict.")
+
         for k, v in iteritems(value):
-            if k not in self.schema:
+            unknown_key = k not in self.schema
+
+            if unknown_key and self._strict:
                 raise InvalidError(
-                    f"{k!r} is not a valid key as per {self.schema!r}",
+                    "%r is not a valid key as per %r" % (k, self.schema),
```
silverdrake11, line 197: .format() is preferable

silverdrake11 (changes requested):

PR #158, opened 2023-04-19
send ubuntu pro reboot output

patriacaelum:
- Adds a monitor plugin to check if a reboot is required
- Update documentation on available plugins


Perfect5th (approved):
LGTM

PR #159, opened 2023-05-09
expand manage_sources_list_d to compensate for poor config parsing

Perfect5th:


landscape/client/manager/aptsources.py
```diff
@@ -132,7 +132,13 @@ def _handle_sources(self, ignored, sources):
             if os.path.isfile(saved_sources):
                 shutil.move(saved_sources, self.SOURCES_LIST)
 
-        if self.registry.config.manage_sources_list_d:
+        if self.registry.config.manage_sources_list_d not in (
```
silverdrake11, line 135: It would be nice if there was a helper function for this, so it could be used in other config parameters

Perfect5th: I'm hesitant to over-engineer this fix, as I hope we can replace our dated config parsing with the more modern approach, which should handle datatypes like this internally.

silverdrake11: What do you think about just a variable in landscape/constants.py like FALSE_VALUES = ["false", "0", "no", ..]

Perfect5th: Hmm yeah that's fine.

silverdrake11 (approved):
LGTM

PR #160, opened 2023-05-16
make manage_sources_list_d conf-file-only

Perfect5th:
There's some really weird interactions going on with the default CLI param value and the value in the config file. To reduce the impact, I've decided that this config will be only in the conf file until we upgrade our configparser and argparser issues.


silverdrake11 (approved):

karlg100:
you should update the man page. it still shows --manage-sources-list-d as a valid option.

PR #161, opened 2023-05-22
Reporter entry point should try to repair 'unsupported locale setting' when detected (LP: #2020064)

juanmitaboada:
This would help to prevent:
https://bugs.launchpad.net/landscape-client/+bug/2020064

Please, check the bug at Launchad before merging.

(The PR includes the proposed solution and test to verify this specific scenario)

### DIFF
--- a/landscape/client/package/reporter.py
+++ b/landscape/client/package/reporter.py
@@ -3,6 +3,7 @@ try:
 except ImportError:
     import urllib.parse as urlparse
 
+import subprocess
 import locale
 import logging
 import time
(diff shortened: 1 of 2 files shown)


Perfect5th:
I don't think I like this solution as-is. First thing that occurs to me: if we are not running as root, or as a password-less sudoer, does this block on `sudo` or through an exception? I think we need appropriate handling/logging for those cases.

Does it possibly make more sense to `locale-gen/update-locale` during installation, say in debian/landscape-client.postinst? Interested in your thoughts on that.

juanmitaboada:
I agree with you. I will address this issue through `postinst` script.

PR #162, opened 2023-05-30
Add SnapMonitor, SnapManager, and related machinery

Perfect5th:


landscape/client/monitor/snapmonitor.py
```diff
@@ -0,0 +1,38 @@
+import logging
+
+from landscape.client.monitor.plugin import DataWatcher
+from landscape.client.snap.http import SnapdHttpException
+from landscape.client.snap.http import SnapHttp
+
+
+class SnapMonitor(DataWatcher):
+
+    run_interval = 1800  # 30 minutes
```
silverdrake11, line 10: Seems like this value is a always overwritten?

landscape/client/monitor/snapmonitor.py
```diff
@@ -0,0 +1,38 @@
...
+        self._snap_http = SnapHttp()
+
+    def register(self, registry):
+        self.config = registry.config
+        self.run_interval = self.config.snap_monitor_interval
+
+        super(SnapMonitor, self).register(registry)
+
+    def get_data(self):
+        return get_installed_snaps(self._snap_http)
```
silverdrake11, line 28: The DataWatcher class will cache the get_data value but since it includes timestamps, it wont so probably best to remove them and sort keys, (even though eventually in the message they get stripped out later).

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,261 @@
...
+                loop.stop()
+                return
+
+            for _ in range(len(change_queue)):
+                cid, name = change_queue.popleft()
+
+                # It's possible (though unlikely) that a change is not in the
+                # list - snapd could have dropped it for some reason. We need
+                # to know if that happens, hence the extra `found` check.
+                found = False
```
silverdrake11, line 170: Use the found flag and breaks in the double loop is a bit confusing. I think it would be best to define result_dict on line 155 which would be the same as result but keyed by id. Then eliminate the second loop and check in there if it exists or not

landscape/client/manager/snapmanager.py
```diff
@@ -0,0 +1,261 @@
...
+
+class SnapManager(ManagerPlugin):
+    """
+    Plugin that updates the state of snaps on this machine, installing,
+    removing, refreshing, enabling, and disabling them in response to messages.
+
+    Changes trigger SnapMonitor to send an updated state message immediately.
+    """
+
+    def register(self, registry):
```
silverdrake11, line 24: Might be more clear to move some initialization into  an _init_ () like self._snap_http and self.SNAP_METHODS

silverdrake11 (changes requested):

landscape/client/monitor/snapmonitor.py
Perfect5th: Right. I think I forgot to delete it when I added the config var. Will remove.

landscape/client/monitor/snapmonitor.py
Perfect5th: Good point. They only get stripped once they are put in the message. I _could_ store the _message_, though. Thinking about it. I will adjust either way.

landscape/client/monitor/snapmonitor.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_installed_snaps(snap_http):
```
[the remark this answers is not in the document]
Perfect5th: I don't think so. snap http raises `SnapdHttpExceptions`, it doesn't handle them. I prefer the logging in the manager/monitor classes. This is only a couple of lines and it is a bit of a stop-gap to get SnapMonitor to execute. I think I can do that more directly in SnapManager. Will try.

landscape/client/manager/snapmanager.py
Perfect5th: I like that. I'll retool it to avoid the extra loop and nesting.

landscape/client/manager/snapmanager.py
Perfect5th: done.

Perfect5th:
Addressed comments above. Please let me know if you want manual testing instructions.

silverdrake11 (approved):

PR #163, opened 2023-06-15
Limit message data directory size

silverdrake11:
Manual testing instructions:

 1. Modify `directory_size` to be 10 instead of 1000 in the class `__init__` of `store.py`
 1. Change the "Sending payload" and "Received payload" to not print the payload in client/broker/transport.py. This makes reading the output logs easier (optional)
 1.  Connect this client to a server and accept the computer
 1. Either stop the server or stop client to make the message system unreachable (change message system config `url` to be something that doesn't exist. Just changing the url suffix should be sufficient)
 1. Start client and take a look at the `messages` directory which is inside of the `data-path` config and watch the directory grow (should not take long).  Once there are 4 folders in there observer the 5th one to get deleted (max_dirs is 4).  Note there may be a slight delay with the deletion since it doesn't happen until after the 5th directory is created. 
 1. To test the deletion by megabytes,  set the `max_size_mb` to something really small like 0.4 kb which is probably the size of the directory already

### DIFF
--- a/landscape/client/broker/store.py
+++ b/landscape/client/broker/store.py
@@ -312,14 +312,14 @@ class MessageStore:
         Delete messages dirs if there's any over the max, which happens if
         messages are queued up but not able to be sent
         """
-            
+
         cur_dirs = os.listdir(self._directory)
         cur_dirs.sort(key=int)  # Since you could have 0, .., 9, 10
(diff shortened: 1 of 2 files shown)


Perfect5th (approved):
+1 LGTM

PR #166, opened 2023-07-13
Added a little check to provide a better error (LP: #2027521)

juanmitaboada:
The logging library should provide a better error using "!r" to avoid confusing messages. 

I have added a little check to provide a better error from our side, now the error would look like this:

`AttributeError: Unknown level "'info'", conversion to logging code was "Level 'INFO'"`

### DIFF
--- a/landscape/lib/logging.py
+++ b/landscape/lib/logging.py
@@ -38,16 +38,22 @@ def init_app_logging(logdir, level="info", progname=None, quiet=False):
     """Given a log dir, set up logging for an application."""
     if progname is None:
         progname = os.path.basename(sys.argv[0])
-    level = logging.getLevelName(level.upper())
-    _init_logging(
-        logging.getLogger(),
-        level,
-        logdir,
-        progname,
-        logging.Formatter(FORMAT),
-        sys.stdout if not quiet else None,
-    )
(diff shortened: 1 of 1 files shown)


Perfect5th:
While this looks good in the case where landscape-client entry points are called directly, e.g., `sudo -u landscape python3 /usr/bin/landscape-broker` I don't think it improves the case where `landscape-config` is called. I built this branch into a .deb, installed it and tried this:

```
$ sudo landscape-config --account-name=devel --computer-title=mytestclient --log-level="'info'"
Please wait...
Failure: twisted.internet.error.ConnectError: An error occurred while connecting: 2: No such file or directory.
```
After a significant timeout.

And there are no logs either in /var/log/landscape, as stated by LP: #2027521.

Raising an exception is great, but we also need some kind of messaging to the landscape-config process that indicates what happened: that landscape-broker didn't start. Maybe log something to an alternate logfile, like `config.log` and we can either report its contents to the user or indicate that they should check there?

Perfect5th (approved):
Nice!

Perfect5th:
Waiting on getting my permissions reinstated (since this repo changed orgs) so I can merge again.

PR #167, opened 2023-07-25
use --is-registered as systemd service start condition

Perfect5th:
This re-resolves https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/2027613 - see the discussion there regarding the rationale behind this change.


Perfect5th:
Will wait to merge until it is approved as a solution in the SRU.

PR #182, opened 2023-09-05
pass ssl_public_key settings into pinger to support self-signed cert TLS

Perfect5th:
To test:
  1. Run landscape server using a self-signed certificate (landscape-scalable charm bundle should give you this by default)
  2. Make sure `ping_url` is `https`
  3. Run client with `--ssl-public-key=<PATH_TO_PUB_KEY.crt>` or set `ssl_public_key = <PATH_TO_PUB_KEY.crt>` in your .conf
  4. Register your client and make sure it can ping. Easiest way to check is to `tail -f /var/log/haproxy.log | grep ping` on your haproxy unit in the juju model. Make sure the traffic is on port `443`.
  
  If you're having trouble getting the cert from your server to your client, a quick way is:
  ```
  echo -n | openssl s_client -connect <HAPROXY_URL>:443 | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' | sudo tee /etc/landscape/server.crt
  ```

Please also make sure this doesn't break plain HTTP pings.

### DIFF
--- a/landscape/client/broker/ping.py
+++ b/landscape/client/broker/ping.py
@@ -51,11 +51,12 @@ from landscape.lib.log import log_failure
 class PingClient:
     """An HTTP client which knows how to talk to the ping server."""
 
-    def __init__(self, reactor, get_page=None):
+    def __init__(self, reactor, get_page=None, cainfo=None):
         if get_page is None:
             get_page = fetch
         self._reactor = reactor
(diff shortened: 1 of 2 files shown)


patriacaelum (approved):
LGTM. It only took me a while to setup juju correctly.

PR #184, opened 2023-09-07
Enable experimental build for strictly-confined snap

Perfect5th:


Perfect5th:
Testing requires building the snap and installing it:
1. `snapcraft`
2. `sudo snap install ./landscape-client_23.08_amd64.snap --dangerous` (or `--devmode`)
3. `sudo landscape-client.config` or edit `/var/snap/landscape-client/common/var/lib/landscape-client.conf` and `sudo start --enable landscape-client`

landscape/client/__init__.py
```diff
@@ -0,0 +1,11 @@
+import os
+
```
silverdrake11, line 2: Another possibility is to have these values in landscape/constants.py

landscape/client/broker/client.py
```diff
@@ -142,11 +142,13 @@ def _run_with_error_log(self):
...
     def _error_log(self, failure):
         """Errback to log and reraise uncaught run errors."""
-        msg = "{} raised an uncaught exception".format(type(self).__name__)
+        cls = type(self).__name__
+        msg = f"{cls} raised an uncaught exception"
         if sys.exc_info() == (None, None, None):
             error(msg)
         else:
             exception(msg)
+        debug(str(failure.value))
```
silverdrake11, line 151: Maybe the traceback module would be useful here?

landscape/client/monitor/ubuntuproinfo.py
```diff
@@ -28,7 +29,25 @@ def get_data(self):
...
+def get_ubuntu_pro_info() -> dict:
+    """Query ua tools for Ubuntu Pro status as JSON, parsing it to a dict.
+
+    If we are running on Ubuntu Core, Pro does not exist - returns a message
+    indicating this.
+    """
+    if IS_CORE:
+        return {
+            "errors": [
+                {
```
silverdrake11, line 41: There's some repetition here between the if and else conditions, could create a function and pass in the error message / parameters, and it would return the json

snap/snapcraft.yaml
```diff
@@ -1,98 +1,110 @@
...
+    environment:
+      LANDSCAPE_CLIENT_SNAP: 1
+      LANDSCAPE_CLIENT_CORE: 1
+      PYTHONPATH: $SNAP/usr/lib/python3/dist-packages:$PYTHONPATH
+
+layout:
+  /etc/landscape-client.conf:
+    bind-file: $SNAP_COMMON/etc/landscape-client.conf
+  /var/lib/landscape/client:
+    bind: $SNAP_DATA/var/lib/landscape/client
```
silverdrake11, line 47: Where is SNAP_DATA? It might be weird if these files are buried in subdirectories where there's nothing else inside

silverdrake11 (approved):
LGTM but left a few comments

landscape/client/__init__.py
Perfect5th: That's a decent thought, but they aren't really constant. They vary based on the env vars.

landscape/client/broker/client.py
Perfect5th: Yup. Good idea. I'll use it.

snap/snapcraft.yaml
Perfect5th: `$SNAP_DATA` is (currently) `/var/snap/landscape-client/<revision>` (https://snapcraft.io/docs/environment-variables#heading--snap-data). It's the suggested location for data that the snap uses that can be backed up and restored across `refresh`es and `revert`s. A lot of the snapcrafting docs examples suggest putting data files here, e.g., https://snapcraft.io/docs/snap-layouts.

landscape/client/monitor/ubuntuproinfo.py
Perfect5th: I'll revisit this.

landscape/client/broker/client.py
Perfect5th: Actually it seems that `traceback` may not play nicely with the `errBack` used in Twisted. Checking it out.

silverdrake11 (approved):

PR #189, opened 2023-10-05
Add optional hostagent_uid field to the message schema.

wck0:
Manual Testing Instructions:

1. Get this PR code.
2. Setup a development `landscape-server` instance.
3. Install this code to the instance running server: `./build/venv/bin/pip install "git+https://github.com/wck0/landscape-client.git"`
4. Run the server code.
5. Edit the `landscape-client.conf` file for `landscape-client` to point to your dev server.
6. Add a new line to that same file: `hostagent_uid = some test value`
7. Run the development client: `sudo ./scripts/landscape-client`
8. Wait for it to register with your dev server.
9. Watch the logs or check in `/tmp/landscape/broker.log` that the registration message includes the `hostagent_uid` key with the value you set above and that it is using version 3.4 of the message schema.
10. Verify you can accept the computer as usual.

### DIFF
--- a/landscape/__init__.py
+++ b/landscape/__init__.py
@@ -22,7 +22,7 @@ DEFAULT_SERVER_API = b"3.2"
 # 3.3:
 #  * Add new schema for the "registration" message, providing Juju information
 #
-SERVER_API = b"3.3"
+SERVER_API = b"3.4"
 
 # The "client-api" field of outgoing messages will be set to this value, and
 # used by the server to know which schema do the message types accepted by the
--- a/landscape/__init__.py
+++ b/landscape/__init__.py
@@ -22,7 +22,7 @@ DEFAULT_SERVER_API = b"3.2"
 # 3.3:
 #  * Add new schema for the "registration" message, providing Juju information
 #
-SERVER_API = b"3.4"
+SERVER_API = b"3.3"
 
 # The "client-api" field of outgoing messages will be set to this value, and
 # used by the server to know which schema do the message types accepted by the
--- a/README
+++ b/README
@@ -81,10 +81,15 @@ system bus.
 $ sudo ./scripts/landscape-client -c root-client.conf
 ```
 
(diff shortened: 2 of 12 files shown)


landscape/__init__.py
```diff
@@ -22,7 +22,7 @@
 # 3.3:
 #  * Add new schema for the "registration" message, providing Juju information
 #
-SERVER_API = b"3.3"
+SERVER_API = b"3.4"
```
silverdrake11, line 25: Since adding an optional field to the registration message, doesn't break anything or change the format of the api, it's not necessary to create a new registration message/ new version of the api.

silverdrake11 (changes requested):

silverdrake11:
What is the user experience like? Will someone have to manually provide the hostagent uid every time they spin up a WSL instance? Or will the windows host somehow set it in the client.conf file?

landscape/__init__.py
wck0: OK, will make that change.

wck0:
> What is the user experience like? Will someone have to manually provide the hostagent uid every time they spin up a WSL instance? Or will the windows host somehow set it in the client.conf file?

As it stands right now, yes, it will need to be manually set in the registry key that serves as the service.conf file. How that user experience will evolve is an open question.

silverdrake11:
Is there anything we can do to obtain this automatically. Maybe it is stored somewhere in the ubuntu file system?

wck0:
> Is there anything we can do to obtain this automatically. Maybe it is stored somewhere in the ubuntu file system?

No, the uid only known by the host machine. It's possible for the hostagent to pass that along to the instance upon creation, or inject it into the registry key / service.conf during a config phase, but that functionality does not yet exist.

silverdrake11 (approved):

silverdrake11:
Tested again and LGTM

PR #190, opened 2023-10-05
fix format of architectures in snapcraft.yaml so launchpad likes it

Perfect5th:
prior to this change builds would only happen on amd64. Here's proof that with this change it builds for all listed arches: https://launchpad.net/~mitchburton/+snap/snapcraft-landscape-client-b8bc866f93dd763be09f288a7a444fec


srunde3 (approved):
LGTM -- saw it working live :smiley_cat:

wck0 (approved):

PR #198, opened 2023-11-25
upstream fix for LP #2040924

Perfect5th:
Upstreaming fix contributed to ubuntu: https://code.launchpad.net/~mitchburton/ubuntu/+source/landscape-client/+git/landscape-client/+merge/456212

bug: https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/2040924

### DIFF
--- a/debian/landscape-common.postinst
+++ b/debian/landscape-common.postinst
@@ -90,6 +90,12 @@ case "$1" in
         WRAPPER=/usr/share/landscape/landscape-sysinfo.wrapper
         PROFILE_LOCATION=/etc/profile.d/50-landscape-sysinfo.sh
         UPDATE_MOTD_LOCATION=/etc/update-motd.d/50-landscape-sysinfo
+        STAMP_LOCATION=/var/lib/landscape/landscape-sysinfo.cache
+
+        touch $STAMP_LOCATION
+        chown landscape:root $STAMP_LOCATION
+        chmod 666 $STAMP_LOCATION
+
         if [ "$RET" = "Cache sysinfo in /etc/motd" ]; then
             rm -f $PROFILE_LOCATION 2>/dev/null || true
             ln -sf $WRAPPER $UPDATE_MOTD_LOCATION
--- a/debian/landscape-common.postinst
+++ b/debian/landscape-common.postinst
@@ -94,7 +94,7 @@ case "$1" in
 
         touch $STAMP_LOCATION
         chown landscape:root $STAMP_LOCATION
-        chmod 666 $STAMP_LOCATION
+        chmod 0644 $STAMP_LOCATION
 
         if [ "$RET" = "Cache sysinfo in /etc/motd" ]; then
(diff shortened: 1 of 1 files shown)


debian/landscape-common.postinst
```diff
@@ -90,6 +90,12 @@ case "$1" in
         WRAPPER=/usr/share/landscape/landscape-sysinfo.wrapper
         PROFILE_LOCATION=/etc/profile.d/50-landscape-sysinfo.sh
         UPDATE_MOTD_LOCATION=/etc/update-motd.d/50-landscape-sysinfo
+        STAMP_LOCATION=/var/lib/landscape/landscape-sysinfo.cache
+
+        touch $STAMP_LOCATION
+        chown landscape:root $STAMP_LOCATION
+        chmod 666 $STAMP_LOCATION
```
panlinux, line 97: Why perms `666`? That's world writable. Didn't you mean `0644`?

Perfect5th: The file is written (or really, replaced) when motd is executed. I'd assumed it would need to be world writable for that to happen. Is that assumption incorrect?

panlinux: The `/etc/update-motd.d/50-landscape-sysinfo` script runs as root at login time. Maybe the original bug was misdiagnosed?

Perfect5th: The original bug was related to the `show-motd` package, which I think is intended to be executed by non-root users as well, but that's just my interpretation based on the bug report and the description of the package (sic):

```
 This package installs a script in /etc/profile.d that dynamically
 generates and shows a message-of-the-day in inteactive shells by
 running scripts installed in /etc/update-motd.d.
```

panlinux: Then why did it start happening just now? Something else changed. But creating a world writable file in `/var/lib` is not the solution. The way it is now, any user can write to it, and that will be displayed on login.

Perfect5th: I don't know why it would suddenly have changed, but I agree with you that permissions would be better off as `0644` here. Changing it.

panlinux: I think we need to understand what is going on, and why this fix or change is even needed.

Perfect5th: I've added a comment to the original bug: https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/2040924/comments/2
If I had the time, I'd go digging myself for the root cause.

panlinux: I troubleshooted it: https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/2040924/comments/3

Perfect5th: I've added a revert of this change and a fix to https://code.launchpad.net/~mitchburton/ubuntu/+source/landscape-client/+git/landscape-client/+merge/454434

Once that's past review, I will contribute the resulting fix here as well.

Perfect5th:
Closing as this was fixed in #201 and #219

PR #199, opened 2023-11-29
fixing inaccuracies with bfree, using bavail instead

amvapor:
The bfree variable (on ubuntu 22.04, my system) Reports a much higher value than bavail, this conceivably occurs do to some blocks being free, but not available due to block sizes or other values, so the disk plugin does not agree with the `df` command, I'm not sure if that consistency is important to this application, but it's important to me, so thought I'd submit a pull request.


amvapor:
Failure on my part, reserved blocks were still in place during my testing, retracting this pr

PR #202, opened 2024-01-16
conditionally use snap_http, fall back to local module

Perfect5th:
Transitional change that allows the [snap-http](https://github.com/Perfect5th/snap-http/) package to be used if it is available. If it's not, previous behaviour is preserved, but now in an API-compatible way.

To test, install `snap-http` from my PPA: https://launchpad.net/~mitchburton/+archive/ubuntu/snap-http
or from pypi.

### DIFF
--- a/landscape/client/manager/snapmanager.py
+++ b/landscape/client/manager/snapmanager.py
@@ -6,10 +6,19 @@ from twisted.internet import task
 from landscape.client.manager.plugin import FAILED
 from landscape.client.manager.plugin import ManagerPlugin
 from landscape.client.manager.plugin import SUCCEEDED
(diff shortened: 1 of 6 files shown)


Perfect5th:
Getting coverage is going to be challenging here (as it's coverage of imports). Might have to ignore the coverage part of the review.

silverdrake11:
What if we copy the module to the repo (like clone from github). It could be for example, a build step

Perfect5th:
> What if we copy the module to the repo (like clone from github). It could be for example, a build step

Might work – could use a git submodule?

Perfect5th:
Closing in favour of #203

PR #203, opened 2024-01-19
Add getting and setting snap config

st3v3nmw:
TL;DR

This PR:
- Adds functionality to set the snap config (via `SnapManager`) & periodically syncing the config (via `SnapMonitor`)
- Migrates Snap HTTP functionality to use the `snap-http` package

---

Deets
----

This started off #202 (_conditionally use snap_http, fall back to local module_) but ran into some issues while using both `snap-http` & the local `landscape.client.snap` simultaneously:
- For HTTP responses, `landscape.client.snap` returns a `dict` while `snap_http` returns `SnapdResponse`
- `snap_http`'s `SnapException` is missing the `json` method in `landscape.client.snap`'s `SnapdHttpException`

Applied the git submodule suggestion in #202 and created a symbolic link: `landscape.client.snap_http -> snap-http/snap_http` and deleted `landscape.client.snap`.

--

<strike>

Other noteworthy change: Added a new "schema type" `Nested` that allows nesting of "container types" like `List`, `Dict`, & `KeyDict`.  The values of snap configs have arbitrary types and can be deeply nested. Consider values like:

```python
# A
{"bar": "default", "baz": False}

# B
{
            "foo": {"baz": "default", "qux": [1, True, 2.0]},
            "bar": "enabled",
}

# C
{
            "foo": {"bar": "baz", "qux": {"quux": "corge"}},
            "ssids": [{"name": "Airport Wi-Fi"}, {"name": "Totally Legit Wi-Fi"}],
}
```

Arbitrary values cannot be generally/succinctly represented with the existing "schema types" esp. very very nested values. The only other option is using a type like `Unknown` that doesn't do any schema validation.

</strike>


Perfect5th:
>Arbitrary values cannot be generally/succinctly represented with the existing "schema types" esp. very very nested values. The only other option is using a type like Unknown that doesn't do any schema validation.

Because the shape of this data is essentially arbitrary, and therefore cannot cleanly map to anything on the Landscape Server side, database-schema-wise, I'm going to suggest that we just use `Unicode` as the type, and serialize it as JSON. Is it guaranteed to be serializable?

There's [some precedence](https://github.com/canonical/landscape-client/blob/e812505b5cb214344a6da91e62ee89f2c2872d8d/landscape/message_schemas/server_bound.py#L352) for doing it this way.

st3v3nmw:
> Is it guaranteed to be serializable?

Yes, the `GET /v2/snaps/{name}/conf` endpoint returns a JSON map of configuration keys and values. Let me switch to `Unicode`

Perfect5th (approved):
LGTM, thanks!

PR #204, opened 2024-01-22
Read temperature from hwmon devices

jck112:
On most systems `hwmon` provides access to the temperature information on more devices, including the CPU and GPU on most systems. These additional temperature sources yield a more accurate maximum temperature for the overall system.

With this change, the following precedence is used for reading thermal zones (depending on availability):
1. `/sys/class/hwmon/*/temp*_input`
2. `/sys/class/thermal/*/temp`
3. `/proc/acpi/thermal_zone/*/temperature`

### DIFF
--- a/landscape/lib/sysstats.py
+++ b/landscape/lib/sysstats.py
@@ -1,3 +1,4 @@
+import glob
 import os.path
 import struct
 import time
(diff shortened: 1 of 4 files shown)


landscape/lib/sysstats.py
```diff
@@ -110,11 +112,20 @@ class ThermalZone:
...
-    def __init__(self, base_path, name):
-        self.name = name
-        self.path = os.path.join(base_path, name)
-        temperature_path = os.path.join(self.path, "temp")
-        if os.path.isfile(temperature_path):
+    def __init__(self, temperature_path):
+        self.path = os.path.dirname(temperature_path)
+        self.name = os.path.basename(self.path)
+        if os.path.basename(temperature_path) == "temperature":
+            for line in open(temperature_path):
```
Perfect5th, line 119: Please use `with open` when opening the file. I realize this code was already like this – this is an opportunity to improve it.

Perfect5th (changes requested):
One minor inline comment. There's also a test that needs to be fixed.

landscape/lib/sysstats.py
jck112: No problem, fixed!

jck112:
@Perfect5th, I fixed the broken test and made the change you requested. Thanks for the review!

Perfect5th:
Thanks for fixing the test! Unfortunately it looks like we're lacking test coverage on a few lines. Do you mind improving that as well?

Perfect5th (approved):
LGTM, thanks much for the contribution!

PR #205, opened 2024-01-25
Add snapd device information to ComputerInfo message

st3v3nmw:
depends on https://github.com/Perfect5th/snap-http/pull/6

### DIFF
--- a/landscape/client/snap_utils.py
+++ b/landscape/client/snap_utils.py
@@ -1,0 +1,38 @@
+import yaml
+
+from landscape.client import snap_http
+from landscape.client.snap_http import SnapdHttpException
+
+
+def get_assertions(assertion_type: str):
+    """Get and parse assertions."""
+    try:
+        response = snap_http.get_assertions(assertion_type)
+    except SnapdHttpException:
+        return
+
+    # the snapd API returns multiple assertions as a stream of
+    # bytes separated by double newlines, something like this:
+    # <assertion1-headers>: <value>
+    # <assertion1-headers>: <value>
+    #
+    # signature
+    #
+    # <assertion2-headers>: <value>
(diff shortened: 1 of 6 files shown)


landscape/client/snap_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    #
+    # <assertion2-headers>: <value>
+    # <assertion2-headers>: <value>
+    #
+    # signature
+
+    # extract the assertion headers + their signatures as separate assertions
+    sections = response.result.decode().split("\n\n")
+    raw_assertions = [
+        "\n\n".join(sections[i : i + 2]) for i in range(0, len(sections), 2)
```
Perfect5th, line 38: This split/re-join could be restructured to be more immediately understandable.

Let's think of it as I/O: you have a bytestream with alternating header groups and signatures separated by `'\n\n'`. These are pairs: `(headers, signature)`. Seems like a perfect usecase for `itertools.bulk`. Too bad that's py3.12 only. Here's an alt:

```python3
sections = response.result.decode().split("\n\n")
parsed = []
rest = sections
while rest:
    headers, signature, *rest = rest
    parsed.append(parse_assertion(headers, signature))

return parsed
```

Let me know what you think.

Perfect5th (changes requested):
I think it needs a bit of fixing, in terms of code clarify. See inline comment.

landscape/client/snap_utils.py
st3v3nmw: Updated to use the more readable code with a slight modification to handle cases where there are no assertions retrieved from the API (`response.result = b""` -> `response.result.decode().split("\n\n")
 = [""]` ).

Perfect5th (approved):
LGTM, thanks!

PR #209, opened 2024-01-31
Remote script execution for snaps

mcw-work:
* Enable remote script execution in the snap.
* When a snap execute all scripts as root (confined by the snap)
* Add LD_LIBRARY and PYTHONPATH env variables to script execution environment.

### DIFF
--- a/landscape/client/manager/scriptexecution.py
+++ b/landscape/client/manager/scriptexecution.py
@@ -99,8 +99,9 @@ class ScriptRunnerMixin:
         # to mention we can't get errno.
         # Don't attempt to change file owner if the client is a snap
         os.chmod(filename, 0o700)
-        if uid is not None or not IS_SNAP:
-            os.chown(filename, uid, gid)
+        if not IS_SNAP:
+            if uid is not None:
+                os.chown(filename, uid, gid)
 
         script = build_script(shell, code)
         script = script.encode("utf-8")
@@ -249,12 +250,14 @@ class ScriptExecutionPlugin(ManagerPlugin, ScriptRunnerMixin):
             full_filename = os.path.join(attachment_dir, filename)
             with open(full_filename, "wb") as attachment:
                 os.chmod(full_filename, 0o600)
-                if uid is not None or not IS_SNAP:
-                    os.chown(full_filename, uid, gid)
+                if not IS_SNAP:
+                    if uid is not None:
+                        os.chown(full_filename, uid, gid)
                 attachment.write(data)
         os.chmod(attachment_dir, 0o700)
-        if uid is not None or not IS_SNAP:
-            os.chown(attachment_dir, uid, gid)
+        if not IS_SNAP:
+            if uid is not None:
+                os.chown(attachment_dir, uid, gid)
         returnValue(attachment_dir)
 
     def run_script(
@@ -300,15 +303,15 @@ class ScriptExecutionPlugin(ManagerPlugin, ScriptRunnerMixin):
             "USER": user or "",
             "HOME": path or "",
         }
-        for locale_var in (
+        for env_var in (
             "LANG",
             "LC_ALL",
             "LC_CTYPE",
             "LD_LIBRARY_PATH",
             "PYTHONPATH",
         ):
-            if locale_var in os.environ:
-                env[locale_var] = os.environ[locale_var]
+            if env_var in os.environ:
+                env[env_var] = os.environ[env_var]
         if server_supplied_env:
             env.update(server_supplied_env)
         old_umask = os.umask(0o022)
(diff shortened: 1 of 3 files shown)


landscape/client/manager/scriptexecution.py
```diff
@@ -173,6 +175,8 @@ def _handle_execute_script(self, message):
         opid = message["operation-id"]
         try:
             user = message["username"]
+            if IS_SNAP:
+                user = "root"
```
Perfect5th, line 179: something about using `root` has really pissed off the github runners. I suggest an additional check for if we're running the tests and not doing this in that case.

Perfect5th (changes requested):

landscape/client/manager/scriptexecution.py
```diff
@@ -296,7 +300,13 @@ def run_script(
...
             "HOME": path or "",
         }
-        for locale_var in ("LANG", "LC_ALL", "LC_CTYPE"):
+        for locale_var in (
+            "LANG",
+            "LC_ALL",
+            "LC_CTYPE",
+            "LD_LIBRARY_PATH",
+            "PYTHONPATH",
+        ):
```
Perfect5th, line 309: the two additions here are not `locale_var`s. I suggest changing the name here.

Perfect5th (changes requested):

mcw-work:
updated tests to match changes to script execution.

mcw-work:
Added unit tests for when script is called from a snap

mcw-work:
Tests all passed

landscape/client/manager/scriptexecution.py
```diff
@@ -96,9 +97,11 @@ def write_script_file(self, script_file, filename, shell, code, uid, gid):
         # It would be nice to use fchown(2) and fchmod(2), but they're not
         # available in python and using it with ctypes is pretty tedious, not
         # to mention we can't get errno.
+        # Don't attempt to change file owner if the client is a snap
         os.chmod(filename, 0o700)
-        if uid is not None:
-            os.chown(filename, uid, gid)
+        if not IS_SNAP:
+            if uid is not None:
```
Perfect5th, line 103: nit: stacked `if`s could be an `and`.

Perfect5th (commented):
Tests pass, nice! Have you (or someone else) ensured that script execution works with this version both on core and classic, then?

mcw-work:
Yeap - I have installed it on a x86 classic machine and a pi core device and they both work fine. The scripts that are saved in the snap-management-demo account on the staging server are also tested and verified as working - that includes one as a heredoc, one using REST API directly and one using Snap-HTTP. Stephen and Yanisa are having a look at what we need to document as well so we don't fall too far behind with that.

landscape/client/manager/tests/test_scriptexecution.py
```diff
@@ -426,6 +437,10 @@ def _run_script(self, username, uid, gid, path):
         patch_chown = mock.patch("os.chown")
         mock_chown = patch_chown.start()
 
+        patch_issnap = mock.patch("landscape.client.IS_SNAP")
+        patch_issnap.return_value = from_snap
+        patch_issnap.start()
```
Perfect5th, line 442: Not sure _exactly_ why this is working, but `return_value` shoudl be set on the return value of `start`, not patch:
```python
patch_issnap = mock.patch("landscape.client.IS_SNAP")
mock_issnap = patch_issnap.start()
mock_issnap.return_value = from_snap
```
I guess the return value isn't being used/checked in any meaningful way in the test?

Perfect5th (changes requested):

landscape/client/manager/tests/test_scriptexecution.py
mcw-work: The return value is just being used to ensure we hit both control paths of the code which we can see is working from the coverage checks. I'll have a go at changing it over and have a rummage through the documentation on patching a global variable and see if there is anything noteworthy.

mcw-work: So, you are right, it wasn't actually working. The patch was not being applied appropriately to an imported global variable - which then overshadowed a different bug in the test. 

To fix that I've tweaked the scriptexecution class so that the import of IS_SNAP is then copied to a local (to the object) variable that the test script can then actually change without needing to patch anything.

mcw-work: Also, as IS_SNAP will default to None, rather than False, the tests should reflect that.

landscape/client/manager/scriptexecution.py
mcw-work: Done

landscape/client/manager/tests/test_scriptexecution.py
Perfect5th: To shed some light on the "patch was not being applied appropriately": `patch` hooks itself into variable _lookup_, not definition, so what you really want to patch is the name of the variable where it's used. In this case, that'd be
```python
mock.patch("landscape.client.manager.scriptexecution.IS_SNAP")
```
totally forgiven for not knowing this little bit of Python errata, and I'm okay with how you're doing it now.

Perfect5th (approved):

PR #210, opened 2024-02-05
add: user management on Core

st3v3nmw:
This change allows us to list, add, and remove SSO users on Ubuntu Core systems.

Example listing on staging:
Screenshot from 2024-02-05 12-51-12

Depends on https://github.com/Perfect5th/snap-http/pull/7

### DIFF
--- a/landscape/client/manager/usermanager.py
+++ b/landscape/client/manager/usermanager.py
@@ -1,10 +1,12 @@
 import logging
 
+from landscape.client import IS_CORE
 from landscape.client.amp import ComponentConnector
 from landscape.client.amp import ComponentPublisher
 from landscape.client.amp import remote
 from landscape.client.manager.plugin import ManagerPlugin
 from landscape.client.monitor.usermonitor import RemoteUserMonitorConnector
+from landscape.client.user.management import SnapdUserManagement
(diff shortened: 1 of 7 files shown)


landscape/client/manager/usermanager.py
```diff
@@ -118,6 +130,15 @@ def _add_user(self, message):
...
         )
 
+    def _add_user_core(self, message):
+        """Run an C{add-user} operation on Core."""
+        return self._management.add_user(
+            message["username"],
+            message["email"],
+            message["sudoer"],
+            message["force-managed"],
+        )
```
Perfect5th, line 140: I don't love this. It's a violation of interface-dependency. I'd prefer it if we modified `add_user` to have a signature that can support both, even if that's just:

    def add_user(self, message):

That way the differences between adding users on core and non-core is contained completely within the `UserManagement` class and doesn't leak up to the `Manager`.

Changing what message types we respond to is _ok_, but alternatively you could simply have the message types we do not respond to be no-ops in `SnapdUserManagement`.

Perfect5th (changes requested):

Perfect5th (approved):
LGTM, thanks!

PR #211, opened 2024-02-07
fix running tasks on the snap manager

st3v3nmw:
These changes were missed during the migration to the `snap-http` library.


Perfect5th (approved):
LGTM, thanks!

PR #212, opened 2024-02-08
Deploy system user assertion to device

st3v3nmw:
Depends on #210

### DIFF
--- a/landscape/client/snap_utils.py
+++ b/landscape/client/snap_utils.py
@@ -1,7 +1,43 @@
+from copy import deepcopy
+
 import yaml
 
 from landscape.client import snap_http
 from landscape.client.snap_http import SnapdHttpException
+
+
+class IgnoreYamlAliasesLoader(yaml.SafeLoader):
+    """Patch `yaml.SafeLoader` to ignore aliases like *alias when loading.
+
+    For instance, a system-user assertion can have the following json:
+      {
+          [...]
+           "system-user-authority": "*",
+          [...]
+      }
+    which after signing gets converted to yaml that looks like:
+      [...]
+      system-user-authority: *
+      [...]
+    pyyaml tries to parse the * as the start of an alias leading to errors.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.yaml_implicit_resolvers = deepcopy(
+            super().yaml_implicit_resolvers,
+        )
+        self.yaml_implicit_resolvers.pop("*", None)
+
+    def fetch_alias(self):
+        return super().fetch_plain()
+
+
+def parse_assertion(headers, signature):
+    """Parse an assertion."""
+    assertion = yaml.load(headers, IgnoreYamlAliasesLoader)
+    assertion["signature"] = signature
+    return assertion
 
 
 def get_assertions(assertion_type: str):
@@ -31,9 +67,8 @@ def get_assertions(assertion_type: str):
         rest = sections
         while rest:
             headers, signature, *rest = rest
-            assertion = yaml.safe_load(headers)
-            assertion["signature"] = signature
-            assertions.append(assertion)
+            parsed = parse_assertion(headers, signature)
+            assertions.append(parsed)
 
     return assertions
 
--- a/landscape/client/user/management.py
+++ b/landscape/client/user/management.py
@@ -3,12 +3,12 @@
 # API, with thorough usage of exceptions and such, instead of pipes to
 # subprocesses. liboobs (i.e. System Tools) is a possibility, and has
 # documentation now in the 2.17 series, but is not wrapped to Python.
-import json
 import logging
 import subprocess
 
 from landscape.client import snap_http
 from landscape.client.snap_http import SnapdHttpException
+from landscape.client.snap_utils import parse_assertion
 from landscape.client.user.provider import UserManagementError
 from landscape.client.user.provider import UserProvider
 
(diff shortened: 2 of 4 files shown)


landscape/client/snap_utils.py
```diff
@@ -4,6 +4,28 @@
...
+    #        "system-user-authority": "*",
+    #       [...]
+    #   }
+    # which after signing gets converted to yaml that looks like:
+    #   [...]
+    #   system-user-authority: *
+    #   [...]
+    # pyyaml tries to parse the * as the start of an alias leading to errors
+    yaml.Loader.fetch_alias = yaml.Loader.fetch_plain
+    yaml.Loader.yaml_implicit_resolvers.pop("*", None)
```
Perfect5th, line 22: This feels like it could have far-reaching side-effects, but I don't think we use yaml (or yaml aliases) in enough other places for it to matter.

landscape/client/snap_utils.py
```diff
@@ -4,6 +4,28 @@
...
+
+    assertion = yaml.load(headers, yaml.Loader)
```
Perfect5th, line 24: Just curious: why can't we use `safe_load` anymore? Is it because the above monkeypatching doesn't work the same on `yaml.SafeLoader`?

landscape/client/user/management.py
```diff
@@ -313,13 +328,25 @@ def add_user(self, message):
...
 
         return response.result
 
+    def _add_system_user_assertion(self, assertion):
+        """Add a system user assertion."""
+        try:
+            # adding an assertion is idempotent
+            snap_http.add_assertion(assertion)
+        except SnapdHttpException as e:
+            result = json.loads(e.args[0])["result"]
```
Perfect5th, line 345: In view or https://github.com/Perfect5th/snap-http/pull/8 being merged can we perhaps use `SnapdHttpException.json` here instead?

Perfect5th (changes requested):

landscape/client/snap_utils.py
st3v3nmw: It actually does work, good catch,... thanks

landscape/client/snap_utils.py
st3v3nmw: I've created a new class `PatchedYamlLoader(yaml.SafeLoader)` with the custom behavior and left `yaml.SafeLoader` untouched. 

For posterity's sake, these are the other attempts:

### Attempt 1

```python
    # load the specific module we want
    loader_spec = importlib.util.find_spec("yaml.loader")
    loader = importlib.util.module_from_spec(loader_spec)
    loader_spec.loader.exec_module(loader)
    # patch it
    loader.SafeLoader.fetch_alias = loader.SafeLoader.fetch_plain
    loader.SafeLoader.yaml_implicit_resolvers.pop("*", None)

    import sys
    # prints: 130478141495936 130478198317968
    print(id(sys.modules["yaml.loader"]), id(sys.modules["yaml"]))

    safe_loader = loader.SafeLoader(headers)
    assertion = safe_loader.get_single_data()
    safe_loader.dispose()
    assertion["signature"] = signature

    import yaml
    # prints: 130478141495936 130478198317968
    print(id(sys.modules["yaml.loader"]), id(sys.modules["yaml"]))

    # this raises the error as before (original behavior without patching)
    yaml.safe_load(headers)
```

If we do this, `yaml.safe_load(headers)` retains its original behavior but the catch is the modules in `sys.modules` both point to the same objects which is suspicious (ids are the same)

### Attempt 2 (reloading)

```python
    # patch
    yaml.SafeLoader.fetch_alias = yaml.SafeLoader.fetch_plain
    yaml.SafeLoader.yaml_implicit_resolvers.pop("*", None)

    import sys
    # prints: 123975180118560 123975179908128
    print(id(sys.modules["yaml.loader"]), id(sys.modules["yaml"]))

    assertion = yaml.load(headers, yaml.SafeLoader)
    assertion["signature"] = signature

    importlib.reload(yaml)
    # prints: 123975180118560 123975179908128
    print(id(sys.modules["yaml.loader"]), id(sys.modules["yaml"]))
    yaml.safe_load(headers)  # original behavior not recovered
```

landscape/client/user/management.py
st3v3nmw: I've made the changes

landscape/client/snap_utils.py
```diff
@@ -4,6 +4,37 @@
...
+    which after signing gets converted to yaml that looks like:
+      [...]
+      system-user-authority: *
+      [...]
+    pyyaml tries to parse the * as the start of an alias leading to errors.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.yaml_implicit_resolvers.pop("*", None)
```
Perfect5th, line 25: I hate to continue to pick on this, but `yaml_implicit_resolvers` seems to be a class attribute on `SafeLoader`, so `pop`ping on it is mutative. See:

```python
>>> import yaml
>>> yaml.SafeLoader.yaml_implicit_resolvers["*"]
[('tag:yaml.org,2002:yaml', re.compile('^(?:!|&|\\*)$'))]
>>> class MyLoader(yaml.SafeLoader):
...     def __init__(self, *args, **kwargs):
...         super().__init__(*args, **kwargs)
...         self.yaml_implicit_resolvers.pop("*", None)
... 
>>> import io
>>> my_loader = MyLoader(io.StringIO())
>>> yaml.SafeLoader.yaml_implicit_resolvers["*"]
  File "<stdin>", line 1, in <module>
KeyError: '*'
```

I think the proper way to exempt this resolver's behaviour is either to override `resolve` (too complex to do here, probably) or override with a copy and mutate that copy.

Perfect5th (changes requested):

landscape/client/snap_utils.py
st3v3nmw: Copying seems to work:

```python
In [1]: import io
   ...: import yaml
   ...: from copy import deepcopy
   ...:

In [2]: yaml.SafeLoader.yaml_implicit_resolvers["*"]
Out[2]: [('tag:yaml.org,2002:yaml', re.compile(r'^(?:!|&|\*)$', re.UNICODE))]

In [3]: class MyLoader(yaml.SafeLoader):
   ...:     def __init__(self, *args, **kwargs):
   ...:         super().__init__(*args, **kwargs)
   ...:         self.yaml_implicit_resolvers = deepcopy(super().yaml_implicit_resolvers)
   ...:         self.yaml_implicit_resolvers.pop("*", None)
   ...:

In [4]: my_loader = MyLoader(io.StringIO())

In [5]: yaml.SafeLoader.yaml_implicit_resolvers["*"]
Out[5]: [('tag:yaml.org,2002:yaml', re.compile(r'^(?:!|&|\*)$', re.UNICODE))]
```

Lemme switch to that.

Perfect5th (approved):
Good work, thanks!

PR #213, opened 2024-02-13
bpickle: guard against negative string/bytestring lengths

Perfect5th:
bpickle has a bug: if you ask it to `loads` a dict, list, or tuple containing a string or bytestring with a negative length, it can enter an infinite loop as it steps back and scans the same section of the string repeatedly.

Fairly minimal reproduction:
```python
from landscape.lib import bpickle

bpickle.loads(b'lu-4test;')
```
This appears to never return.

Fix is to guard against negative length values. Zero values are okay, as you already get a different error for those, though it could perhaps be clearer.


silverdrake11:
LGTM

silverdrake11 (approved):

PR #214, opened 2024-02-14
add: snap services management

st3v3nmw:
In addition to the running processes, we should send the running snap services to the Landscape server. This will then permit the display of these services along with the option to trigger start/stop/restart commands.


Perfect5th (approved):
LGTM thanks!

PR #215, opened 2024-02-15
Fix Livepatch status reporting bug

srunde3:
Fix issue in which `pro status` inaccurately displays Livepatch status as "disabled" even if it is active.  Bug is due to the `landscape` user not being able to run snaps properly since it has a `$HOME` directory outside of `/home` (it's in `/var/lib/`).  

See https://bugs.launchpad.net/landscape-client/+bug/2037670

Manual Testing:
- Spin up a VM with a Livepatch-supported kernel
- Modify `UbuntuProInfo.run_interval` to be something quick like 10
- Modify `landscape-client.conf to point to a Landscape Server instance.  Additionally include `exchange_interval = 10` to exchange messages with server more quickly
- sudo scripts/landscape-monitor --config landscape-client.conf 
- Toggle livepatch and ensure updates are reflected in Server `sudo pro <enable/disable> livepatch`


silverdrake11:
According to the this comment in `get_data()`

```
New messages will only be sent
    when the result of get_data() has changed since the last time it
    was called.
```

However I believe it only works for monitor DataWatcher. If it works this would cut down traffic from once every 15 minutes to once a day or less. I think it's worth caching it, but I don't think it's currently working. See `_persist` in `landscape/client/monitor/plugin.py` for how this works. I'm not sure if the best way to do it is to make a new class, make existing one work, or just define a function in ubuntupro

silverdrake11:
Why was `net-tools` added as a dependency?

Perfect5th:
> Why was `net-tools` added as a dependency?

It's required for some of the unit tests to pass. It's a build dep in `debian/control`. I'm not sure why it's needed in `DEB_REQUIRED` though.

srunde3:
That's my mistake then -- it should go in `debian/control` instead of `DEB_REQUIRED`.  I added it because it was required for tests to pass

srunde3:
Moved `net-tools` dependency into Makefile instead of `setup_lib::DEB_REQUIRED`

Perfect5th (approved):
LGTM, may want to wait for @silverdrake11 to re-review, though.

silverdrake11:
LGTM except for the name `data` that you are persisting to. It would be better to have it be more specific name since this is only used for ubuntu pro manager and not other kinds of data

srunde3:
Discussed offline with @silverdrake11 -- fine to use `data` since persist should be unique to `UbuntuProInfo` based on file path.

PR #217, opened 2024-02-23
refactor: move snap services to separate manager & monitor

st3v3nmw:
~Makes sure the extra `services` field doesn't trip up the server after the next landscape server release~

### DIFF
--- a/landscape/client/monitor/snapservicesmonitor.py
+++ b/landscape/client/monitor/snapservicesmonitor.py
@@ -1,0 +1,28 @@
+import logging
+
+from landscape.client import snap_http
+from landscape.client.monitor.plugin import DataWatcher
+from landscape.client.snap_http import SnapdHttpException
+
+
+class SnapServicesMonitor(DataWatcher):
+
+    message_type = "snap-services"
+    message_key = "services"
+    persist_name = message_type
+    scope = "snaps"
+
+    def register(self, registry):
+        self.config = registry.config
+        self.run_interval = 60  # 1 minute
+        super().register(registry)
+
+    def get_data(self):
+        try:
+            services = snap_http.get_apps(services_only=True).result
+        except SnapdHttpException as e:
+            logging.warning(f"Unable to list services: {e}")
+            services = []
+        services.sort(key=lambda x: x["name"])
+
+        return {"running": services}
--- a/landscape/message_schemas/server_bound.py
+++ b/landscape/message_schemas/server_bound.py
@@ -803,7 +803,17 @@ SNAPS = Message(
                         ],
                     ),
                 ),
-                "services": List(
+            },
+        ),
+    },
+)
+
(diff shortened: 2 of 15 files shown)


Perfect5th:
We need to be a bit careful here. Releasing this client version without backporting the fix to still-supported server versions will break those server versions.

I'd like to suggest a non-backward-breaking option instead: we use a different message type for snap services. Server automatically ignores message types it isn't aware of, and that way reporting installed snaps will still work even if reporting services doesn't.

The new message type can still be sent by `SnapMonitor`, or we could introduce a separate monitor/manager for services, which may give us flexibility if we find that we need service-specific reporting cadence or data handling.

st3v3nmw:
I've refactored the code to introduce a separate manager and monitor set for snap services (`SnapServicesManager` & `SnapServicesMonitor`). 

Also fixed the snaps message in `SnapManager._send_snap_update` to match the outgoing message schema. The list of snaps was being attached to snaps instead of `snaps.installed`. This also matches the behavior of `SnapMonitor`

landscape/client/monitor/snapservicesmonitor.py
```diff
@@ -0,0 +1,27 @@
...
+class SnapServicesMonitor(DataWatcher):
+
+    message_type = "snap-services"
+    message_key = "services"
+    persist_name = message_type
+    scope = "snaps"
+
+    def register(self, registry):
+        self.config = registry.config
+        super().register(registry)
```
Perfect5th, line 17: you may need to set `self.run_interval` for the monitor to be invoked periodically, but I guess until we have a server version that accepts these messages it's kind of moot.

landscape/message_schemas/server_bound.py
```diff
@@ -830,7 +840,6 @@
                     ),
                 ),
             },
```
Perfect5th, line 842: To save us in future of something like this happening again, it might be useful to think whether we can anticipate needing to add new keys to any of these `KeyDict`s, and if so, set them to `strict=False`. If there's no reason to do so for any of these, just let me know.

Perfect5th (changes requested):

landscape/message_schemas/server_bound.py
st3v3nmw: I think it's safe to since the `snap-services` message isn't mature yet.

The inner one `services.running` does have `strict=False`, let me add that to `services`

landscape/client/monitor/snapservicesmonitor.py
st3v3nmw: Added a 60 seconds `run_interval`.

Also, seems like the monitors inherit a default 5 seconds `run_interval` from `DataWatcher > MonitorPlugin > BrokerClientPlugin.run_interval`

Perfect5th:
Also tested against current version of landscape-server. Seems good.

PR #220, opened 2024-02-27
fix: access to files needed for user management

st3v3nmw:
## TL;DR
This change grants write permissions to files, folders, and interfaces required for user management to work on the snap when it's running on Ubuntu classic.

---

## DETAILS
After #218, I realized I was testing in `devmode` where permissions are lax. After switching to `dangerous` which is strict, I realized that nothing works 😞. This PR should get landscape-client snap's user management working on Ubuntu classic.

Here's a laundry list of the issues I played whack-a-mole with along the way (it should also justify choices made with the fixes):

### 1. `adduser` overwriting `$PATH`

Trying to run `adduser` inside the snap's shell (essentially what `landscape.client.user.management.UserManagement` does) led to the following error:

``` console
$ adduser test --debug
Adding user `test' ...
Selecting UID from range 1000 to 59999 ...
Selecting GID from range 1000 to 59999 ...
Adding new group `test' (1001) ...
/sbin/groupadd -g 1001 test
adduser: `/sbin/groupadd -g 1001 test' returned error code 72057594037927935. Exiting.
```

The issue here is we're trying to use `/sbin/groupadd` which we don't have access to:

```
$ /sbin/groupadd
bash: /sbin/groupadd: Permission denied
```

Though we already have groupadd inside the snap (from the staged `passwd` package):

```
$ which groupadd
/snap/landscape-client/x14/usr/sbin/groupadd
$ which adduser
/snap/landscape-client/x14/usr/sbin/adduser
$ echo $PATH
/snap/landscape-client/x14/usr/sbin:/snap/landscape-client/x14/usr/bin:/snap/landscape-client/x14/sbin:/snap/landscape-client/x14/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
```

Digging deeper, we find that the adduser package modifies `$PATH` [here](https://git.launchpad.net/ubuntu/+source/adduser/tree/adduser#n202):

```perl
# explicitly set PATH, because super (1) cleans up the path and makes adduser unusable;
# this is also a good idea for sudo (which doesn't clean up)
$ENV{"PATH"}="/bin:/usr/bin:/sbin:/usr/sbin";
```

Then it runs `which` with the modified `$PATH` [here](https://git.launchpad.net/ubuntu/+source/adduser/tree/adduser#n374):

```perl
log_info( mtx("Adding group `%s' (GID %d) ..."), $new_name, $gid_option);
my $groupadd = &which('groupadd');
```

The fix here is to delete the offending line while building the snap:

```bash
sed -i '\#/bin:/usr/bin:/sbin:/usr/sbin#d' $SNAPCRAFT_PART_INSTALL/usr/sbin/adduser
```

### 2. Access to the kernel's audit interface

Next, we find that `groupadd` can't open the kernel's audit interface:

```console
$ adduser test --debug
Adding user `test' ...
Selecting UID from range 1000 to 59999 ...
Selecting GID from range 1000 to 59999 ...
Adding new group `test' (1001) ...
/snap/landscape-client/x22/usr/sbin/groupadd -g 1001 test
Cannot open audit interface - aborting.
adduser: `/snap/landscape-client/x22/usr/sbin/groupadd -g 1001 test' returned error code 1. Exiting.
```

Fixed by adding the [netlink-audit](https://snapcraft.io/docs/netlink-audit-interface) interface.

### 3. Read & write access to `/etc`

Next, we find that `groupadd` can't create the `/etc/group.lock` lock file to prevent simultaneous updates to `/etc/group`.

```console
$ adduser test --debug
Adding user `test' ...
Selecting UID from range 1000 to 59999 ...
Selecting GID from range 1000 to 59999 ...
Adding new group `test' (1001) ...
/snap/landscape-client/x23/usr/sbin/groupadd -g 1001 test
groupadd: cannot lock /etc/group; try again later.
adduser: `/snap/landscape-client/x23/usr/sbin/groupadd -g 1001 test' returned error code 10. Exiting.
```

Tried fixing this by plugging the [system-files](https://snapcraft.io/docs/system-files-interface) interface with read & write access to the following files:

```
- /etc/.pwd.lock
- /etc/shadow
- /etc/shadow-
- /etc/shadow.lock
- /etc/passwd
- /etc/passwd-
- /etc/passwd.lock
- /etc/pam.d/passwd
- /etc/subuid
- /etc/subuid-
- /etc/adduser.conf
- /etc/default/useradd
- /etc/deluser.conf
- /etc/group
- /etc/group-
- /etc/group.lock
- /etc/gshadow
- /etc/gshadow-
- /etc/gshadow.lock
- /etc/login.defs
- /etc/subgid
- /etc/subgid-
- /etc/skel
```

And it still doesn't work since `groupadd` keeps creating files with random suffixes `/etc/group.N`:

```console
$ adduser test --debug
Adding user `test' ...
Selecting UID from range 1000 to 59999 ...
Selecting GID from range 1000 to 59999 ...
Adding new group `test' (1001) ...
/snap/landscape-client/x25/usr/sbin/groupadd -g 1001 test
groupadd: /etc/group.716838: Permission denied
groupadd: cannot lock /etc/group; try again later.
adduser: `/snap/landscape-client/x25/usr/sbin/groupadd -g 1001 test' returned error code 10. Exiting.

$ adduser test --debug
Adding user `test' ...
Selecting UID from range 1000 to 59999 ...
Selecting GID from range 1000 to 59999 ...
Adding new group `test' (1001) ...
/snap/landscape-client/x25/usr/sbin/groupadd -g 1001 test
groupadd: /etc/group.717005: Permission denied
groupadd: cannot lock /etc/group; try again later.
adduser: `/snap/landscape-client/x25/usr/sbin/groupadd -g 1001 test' returned error code 10. Exiting.
```

We also can't use patterns like `/etc/group*` with `systems-files` as they're not supported:

```
$ sudo snap install landscape-client_23.08_amd64.snap --dangerous
2024-02-28T18:27:33+03:00 INFO snap "landscape-client" has bad plugs or slots: landscape-client-user-management (cannot add system-files plug: "/etc/group*" contains a reserved apparmor char
from ?*[]{}^")
landscape-client 23.08 installed
```

So the fix here is to give read & write access to the entire `/etc` folder. But it still doesn't work because the folder structure is all messed up inside the snap since we're mounting `/etc` with `system-files` and also creating a virtual file system `/etc` with `layout`:

```yaml
[...]
plugs:
  landscape-client-user-management:
    interface: system-files
    write:
      - /etc
[...]

layout:
  /etc/landscape-client.conf:
    bind-file: $SNAP_COMMON/etc/landscape-client.conf
[...]
```

```console
$ df -h
[...]
tmpfs                      7.8G     0  7.8G   0% /etc
[...]
```

Which means that changes made to files like `/etc/group` do not appear on the host's file system. Getting rid of the `/etc/landscape-client.conf` layout fixes this.

### 4. Permission issues writing to `/etc/shadow` & `/etc/gshadow`

At this point we can write to `/etc/group` but can't write to `/etc/gshadow` which means that adding users isn't yet working. If we check the `journalctl` logs, we see:

```
group added to /etc/group: name=testing, GID=1002
SECCOMP auid=1000 uid=0 gid=0 ses=3 subj=snap.landscape-client.landscape-client pid=1090645 comm="groupadd" exe="/snap/landscape-client/x5/usr/sbin/groupadd" sig=0 arch=c000003e syscall=93 compat=0 ip=0x7f1fe2e94c4b code=0x50000
groupadd[1090645]: failed to add group testing to /etc/gshadowkernel: audit: type=1326 audit(1709209456.168:46723): auid=1000 uid=0 gid=0 ses=3 subj=snap.landscape-client.landscape-client pid=1090645 comm="groupadd" exe="/snap/landscape-client/x5/usr/sbin/groupadd" sig=0 arch=c000003e syscall=93 compat=0 ip=0x7f1fe2e94c4b code=0x50000
terra groupadd[1090645]: failed to add group testing
```

In the `SECCOMP` log, we find that we're attempting to do syscall 93 which is [fchown](https://filippo.io/linux-syscall-table/) which is being blocked by seccomp/AppArmor. We can confirm this by checking landscape client's AppArmor profile and see that the `chown` capability is not enabled/set:

```
$  cat /var/lib/snapd/apparmor/profiles/snap.landscape-client.landscape-client | grep chown
/{,usr/}bin/chown ixr,
```

Luckily, the [account control](https://snapcraft.io/docs/account-control-interface) interface grants us this capability:

```
# Capabilities needed by useradd
capability audit_write,
capability chown,
capability fsetid,
```
_from [snapd/interfaces/builtin/account_control.go#L99](https://github.com/snapcore/snapd/blob/5455753b469ad662101afe64fc5e7438f78bb1a6/interfaces/builtin/account_control.go#L99)_

Which also updates our seccomp profile (`cat /var/lib/snapd/seccomp/bpf/snap.landscape-client.landscape-client.src`):
```
# useradd requires chowning to 0:'42'
fchown - u:root 42
fchown32 - u:root 42
```

### 5. Can't create home directories for new users

Next, we realize that we can't create home directories for new users:

```console
$ adduser testing
Adding user `testing' ...
Adding new group `testing' (1001) ...
Adding new user `testing' (1001) with group `testing' ...
Creating home directory `/home/testing' ...
Stopped: Couldn't create home directory `/home/testing': Permission denied.

Removing directory `/home/testing' ...
Removing user `testing' ...
Removing group `testing' ...
groupdel: group 'testing' does not exist
adduser: `groupdel testing' returned error code 6. Exiting.
```

Adding the `home` interface still leads to the same error. It seems like we don't have write access even though the folder is loaded as `read-write`?

```console
$ ls -lah /home
total 16K
drwxr-xr-x  5 root    root    4.0K Feb 29 13:44 .
drwxr-xr-x 21 root    root     580 Feb 29 12:35 ..
[...]
drwxr-x---  2    1001    1001 4.0K Feb 29 13:44 testing
$ rm -rf /home/testing
rm: cannot remove '/home/testing': Permission denied
$ mkdir /home/new
mkdir: cannot create directory ‘/home/new’: Permission denied
$ cat /proc/mounts | grep home
/dev/mapper/vgubuntu-root /home ext4 rw,relatime,errors=remount-ro 0 0
```

If we add `/home` to `system-files` it still fails due to `chown` but we're able to create the user's home directory:

```console
$ adduser testing
Adding user `testing' ...
Adding new group `testing' (1001) ...
Adding new user `testing' (1001) with group `testing' ...
Creating home directory `/home/testing' ...
Stopped: chown 1001:1001 /home/testing: Operation not permitted

Removing directory `/home/testing' ...
Removing user `testing' ...
Removing group `testing' ...
groupdel: group 'testing' does not exist
adduser: `groupdel testing' returned error code 6. Exiting.
```

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -25,5 +25,5 @@ override_dh_auto_install:
 override_dh_installsystemd:
 	dh_installsystemd
 
-override_dh_auto_test:
-	HOME=$(shell mktemp -d) && make check
+# override_dh_auto_test:
+# 	HOME=$(shell mktemp -d) && make check
--- a/landscape/client/__init__.py
+++ b/landscape/client/__init__.py
@@ -7,5 +7,7 @@ USER = "root" if IS_SNAP else "landscape"
 GROUP = "root" if IS_SNAP else "landscape"
 
 DEFAULT_CONFIG = (
-    "/etc/landscape-client.conf" if IS_SNAP else "/etc/landscape/client.conf"
+    os.environ["SNAP_COMMON"] + "/etc/landscape-client.conf"
+    if IS_SNAP
+    else "/etc/landscape/client.conf"
 )
--- a/landscape/client/manager/tests/test_usermanager.py
+++ b/landscape/client/manager/tests/test_usermanager.py
@@ -30,6 +30,7 @@ psmith:!:13348:0:99999:7:::
 sbarnes:$1$q7sz09uw$q.A3526M/SHu8vUb.Jo1A/:13349:0:99999:7:::
 """,
         )
+        self.empty_shadow_file = self.makeFile("")
         accepted_types = ["operation-result", "users"]
         self.broker_service.message_store.set_accepted_types(accepted_types)
         self.plugins = []
@@ -103,7 +104,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
                 ],
             )
 
-        self.setup_environment([], [], None)
+        self.setup_environment([], [], self.empty_shadow_file)
 
         result = self.manager.dispatch_message(
             {
@@ -161,7 +162,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
                 ],
             )
 
-        self.setup_environment([], [], None)
+        self.setup_environment([], [], self.empty_shadow_file)
 
         result = self.manager.dispatch_message(
             {
@@ -220,7 +221,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
                 ],
             )
 
-        self.setup_environment([], [], None)
+        self.setup_environment([], [], self.empty_shadow_file)
 
         result = self.manager.dispatch_message(
             {
@@ -263,7 +264,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
                 ],
             )
 
-        self.setup_environment([], [], None)
+        self.setup_environment([], [], self.empty_shadow_file)
 
         result = self.manager.dispatch_message(
             {
@@ -298,7 +299,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
             self.assertEqual(messages, new_messages)
             return result
 
-        plugin = self.setup_environment([], [], None)
+        plugin = self.setup_environment([], [], self.empty_shadow_file)
         result = self.manager.dispatch_message(
             {
                 "username": "jdoe",
@@ -367,7 +368,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
             )
 
         users = [("bo", "x", 1000, 1000, "Bo,,,,", "/home/bo", "/bin/zsh")]
-        self.setup_environment(users, [], None)
+        self.setup_environment(users, [], self.empty_shadow_file)
         result = self.manager.dispatch_message(
             {
                 "username": "jdoe",
@@ -489,7 +490,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
             ("jdoe", "x", 1001, 1000, "John Doe,,,,", "/home/bo", "/bin/zsh"),
         ]
         groups = [("users", "x", 1001, [])]
-        self.setup_environment(users, groups, None)
+        self.setup_environment(users, groups, self.empty_shadow_file)
         result = self.manager.dispatch_message(
             {
                 "uid": 1001,
@@ -530,7 +531,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
         users = [
             ("jdoe", "x", 1000, 1000, "John Doe,,,,", "/home/bo", "/bin/zsh"),
         ]
-        plugin = self.setup_environment(users, [], None)
+        plugin = self.setup_environment(users, [], self.empty_shadow_file)
         result = self.manager.dispatch_message(
             {
                 "username": "jdoe",
@@ -601,7 +602,7 @@ class UserOperationsMessagingTest(UserGroupTestBase):
             ("jdoe", "x", 1000, 1000, "John Doe,,,,", "/home/bo", "/bin/zsh"),
         ]
         groups = [("users", "x", 1001, ["jdoe"])]
-        self.setup_environment(users, groups, None)
+        self.setup_environment(users, groups, self.empty_shadow_file)
         result = self.manager.dispatch_message(
             {
                 "username": "jdoe",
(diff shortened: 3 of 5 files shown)


st3v3nmw:
Looks like it's impossible to get around that seccomp violation. `snappy-debug`'s only solution is to modify the code not to `chown` but that doesn't seem to be an option here:

```
= Seccomp =
Time: Mar  1 09:26:20
Log: auid=1000 uid=0 gid=0 ses=3 subj=snap.landscape-client.landscape-client pid=23471 comm="adduser" exe="/usr/bin/perl" sig=0 arch=c000003e 92(chown) compat=0 ip=0x7f31a69d1c1b code=0x50000
Syscall: chown
Suggestions:
* don't copy ownership of files (eg, use 'cp -r --preserve=mode' instead of 'cp -a')
* try the snapcraft preload plugin: https://github.com/sergiusens/snapcraft-preload
* adjust program to not use 'chown'
* ignore the denial if the program otherwise works correctly (unconditional chown is often just noise)
```

Closing.

PR #222, opened 2024-02-29
fix(expandvars): replace bash shell expansion with custom routine

st3v3nmw:
The previous shell expansion left us vulnerable to code injection attacks. This change should address that.


Perfect5th (approved):
LGTM, thanks! Apologies again for the back-and-forth on this function.

PR #223, opened 2024-03-01
fix(UserManager): pick correct shadow file on Core

st3v3nmw:
`UserManager` was defaulting to `/etc/passwd` when the class is initialized without any params.


Perfect5th (approved):
LGTM

PR #224, opened 2024-03-04
fix(build snap): switch to the python build plugin

st3v3nmw:
This PR should switch us over to using the `python` build plugin when building the snap to stop depending and being affected by changes on the `.deb` build process.

Make sure to run `snapcraft clean` before doing a manual test


Perfect5th (approved):
Changes to `snapcraft.yaml` LGTM. Just running some manual build/install tests.

PR #226, opened 2024-03-05
fix(ProcessKiller): add process-control interface to the snap

st3v3nmw:
This will allow us to kill and end processes through the snap.

The before & after:

image


Perfect5th (approved):
LGTM, with one follow-up action. Does this plug need permission to be auto-connected? If so, please make a request on the snapstore discord to have that happen.

PR #228, opened 2024-03-07
chore(snap make targets): remove hardcoded snap version

st3v3nmw:
The snap version for the `snap-install` & `snap-clean` make targets is fixed to `0.1`. This change will ensure that it changes dynamically as the `snapcraft.yaml` file evolves.

Developers are required to install [yq](https://github.com/mikefarah/yq) on their machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,8 @@ etags:
 	-etags --languages=python -R .
 
 snap-install:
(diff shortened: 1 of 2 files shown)


Perfect5th (approved):
LGTM, thanks!

PR #230, opened 2024-03-15
Permit Core devices to be managed by Landscape

srunde3:
Include Ubuntu Pro info such that Core devices may register with Landscape under a full Ubuntu Pro license.

To test manually, either spin up a Core instance or set `IS_CORE` in `landscape.client.__init__` to `True`.  Register with server and confirm that the client may be accepted and placed into an Ubuntu Pro full license.

### DIFF
--- a/landscape/client/broker/tests/test_registration.py
+++ b/landscape/client/broker/tests/test_registration.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import socket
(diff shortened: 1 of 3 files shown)


Perfect5th (approved):
LGTM

st3v3nmw (approved):
LGTM as well, thanks.

PR #232, opened 2024-03-22
Add network-manager and network-control plugs

mcw-work:
Added network-manager and network-control plugs to the snap. This is for the D'crypt edge case and will need to be manually connected. We do not have any plans to ask for auto-connection at this stage.


Perfect5th (approved):
LGTM

PR #233, opened 2024-03-22
allow changing the landscape snap config using snap set

st3v3nmw:
This will allow us to change the `landscape-client.conf` config file using snap set.

`snap set` doesn't allow us to view the previous value of the config so we have to clean up the snap config to detect changes on the next run. Unsetting the values helps us avoid conflicts with configuration written through `landscape-client.config`.

It also expands the range of options that can be set. For instance, it's now possible to configure the plugins when using a gadget snap.

Please note that the landscape-client service needs to be restarted after config changes.

Issue reported by @dilyn-corner

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,3 +1,52 @@
-#!/bin/sh -e
+#!/usr/bin/env python3
 
-# Do nothing for now
+import subprocess
+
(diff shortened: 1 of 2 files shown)


Perfect5th (approved):
LGTM, will merge.

PR #234, opened 2024-03-26
Fix registration for snap

srunde3:
The snap cannot call the `pro` client to get Ubuntu Pro info, which occurs during registration.  We still need ubuntu_pro_info to be called during registration to provide mocked pro info for Core (i.e., Core snap) management.

https://bugs.launchpad.net/landscape-client/+bug/2059136


wck0 (approved):
LGTM +1

st3v3nmw (approved):
+1

PR #235, opened 2024-03-27
d/landscape-sysinfo.wrapper fix case on CORES variable

Perfect5th:
Incorrect reference to the `CORES` variable meant that `THRESHOLD` was always `1.0`. Not entirely fatal to the sysinfo script, but does significantly change the odds whether the system is deemed to have too high a load to collect sysinfo.


st3v3nmw (approved):
LGTM

PR #237, opened 2024-04-11
Update snapcraft.yaml to grade:stable to permit snap promotion.

mcw-work:
This change is needed to allow us to promote the snap to the latest stable channel.


st3v3nmw (approved):
LGTM

PR #238, opened 2024-04-18
chore: update remote for snap-http submodule

st3v3nmw:
Noticed that `snap-http` was moved to https://github.com/canonical/snap-http hence the update.

`checkout` on CI okay:
image


Perfect5th (approved):
Yes good catch, thanks!

PR #240, opened 2024-04-26
Fix setlocale for reporter.py

moonsoftsrl:
landscape-package-reporter did not work on ubuntu 22.04 and 20.04 lxc containers provided by proxmox.

The execution of the setlocal function throw this error
```python
>>> locale.setlocale(locale.LC_CTYPE, ("C", "UTF-8"))
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/locale.py", line 608, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
```

This change let the execution be compatible on both vm and lxc with Ubuntu 20.04 and 22.04

### DIFF
--- a/landscape/client/package/reporter.py
+++ b/landscape/client/package/reporter.py
@@ -942,7 +942,7 @@ class FakeReporter(PackageReporter):
 def main(args):
     # Force UTF-8 encoding only for the reporter, thus allowing libapt-pkg to
     # return unmangled descriptions.
-    locale.setlocale(locale.LC_CTYPE, ("C", "UTF-8"))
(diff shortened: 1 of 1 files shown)


srunde3:
Hi there,

Thank you for your PR.  We need to continue to use UTF-8 encoding for internationalization support; it sounds like this is an issue with the environments you're running Landscape Client in.  You should be able to generate additional locales in your containers if they don't have `C.UTF-8` by default, e.g., `sudo locale-gen C.UTF-8`

I'm going to close this PR since I believe this is an environment issue; please feel free to re-open with some additional info if you feel strongly that this should be solved with a code change.

PR #241, opened 2024-04-29
Add non-filtered interfaces to the API response (LP: #2052834)

ghadi-rahme:
Currently not all interfaces show up in the API response (interfaces with no IP address, in the down state, etc...). 
This patch makes any interface that is not filtered out and has a valid MAC address show up in the API response. All previously included interfaces will still show up as expected.

Mainly interfaces in the down state as well as interfaces in the up state but with no IP address assigned to them will now show up in the API response as long as they have a MAC address. 
The IP address, broadcast address and net-mask for these interfaces will be set to `0.0.0.0`.

### DIFF
--- a/landscape/lib/tests/test_network.py
+++ b/landscape/lib/tests/test_network.py
@@ -46,8 +46,12 @@ class NetworkInfoTest(BaseTestCase):
         }
 
         for device in device_info:
-            if device["mac_address"] == "00:00:00:00:00:00":
+            if (
+                device["mac_address"] == "00:00:00:00:00:00"
+                or device["ip_address"] == "0.0.0.0"
+            ):
                 continue
+            print("result is: ", result)
             self.assertIn(device["interface"], result)
             block = interface_blocks[device["interface"]]
             self.assertIn(device["netmask"], block)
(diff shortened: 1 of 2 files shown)


landscape/lib/tests/test_network.py
```diff
@@ -46,8 +46,12 @@ def test_get_active_device_info(self, mock_get_network_interface_speed):
         }
 
         for device in device_info:
-            if device["mac_address"] == "00:00:00:00:00:00":
+            if (
+                device["mac_address"] == "00:00:00:00:00:00"
+                or device["ip_address"] == "0.0.0.0"
+            ):
                 continue
+            print("result is: ", result)
```
srunde3, line 54: Please remove this print from the tests

snap-http
srunde3, line 1: We'll want to make sure this submodule is up-to-date prior to the merge

landscape/lib/tests/test_network.py
```diff
@@ -266,7 +282,7 @@ def test_skip_iface_with_no_ip(self, mock_interfaces, mock_ifaddresses):
     @patch("landscape.lib.network.get_flags")
     @patch("landscape.lib.network.netifaces.ifaddresses")
     @patch("landscape.lib.network.netifaces.interfaces")
-    def test_skip_iface_down(
+    def test_iface_down(
```
srunde3, line 285: Since we're already making edits to this section of the code, would you mind adding a docstring to the test to document this new behavior?  Something like "Interfaces in the 'down' state are still reported"

srunde3 (changes requested):

ghadi-rahme:
Thank you for the review @srunde3!

I added the docstring and removed the print statement, it was a debug remnant.

Let me know if anything else needs changing.

srunde3 (approved):
LGTM.  Thank you for your contribution!

ghadi-rahme:
Thank you for the review and for merging the patch!

Yes you are correct these interfaces still do no show up in the UI, my goal was for them to show up in the API requests.

PR #243, opened 2024-05-06
Added flag to skip sending a registration request and minor fixes

silverdrake11:
When in interactive mode, the user is prompted if a registration message needs to be sent. However in silent mode, no such option exists, and one is always sent. This MR adds a flag to skip the registration request. This has uses such as interacting with WSL and also the client charm.

In addition some minor fixes were made:

- Remove http:// and slashes from landscape domain if user accidentally provides it
- Removed depreciated iteritems to prevent warning in tests
- Closed opened file handles in tests to prevent warning


landscape/client/configuration.py
```diff
@@ -473,7 +480,9 @@ def query_landscape_edition(self):
             self.landscape_domain = self.prompt_get_input(
                 "Landscape Domain: ",
                 True,
-            )
+            ).strip('/')
+            self.landscape_domain = re.sub(r'^https?://', '',
+                                           self.landscape_domain)
```
wck0, line 485: I notice you are using single quotes instead of double quotes like we have elsewhere in the code. Our linters don't catch this, but it would be good to be consistent.

landscape/client/tests/test_configuration.py
```diff
@@ -760,7 +809,9 @@ def get_content(self, config):
         try:
             config.config = config_file
             config.write()
-            return open(config.config).read().strip() + "\n"
+            with open(config.config) as fh:
+                text = fh.read().strip() + "\n"
+            return text
```
wck0, line 814: Thank you for this change.

wck0 (changes requested):
Minor nit that probably should be fixed, otherwise LGTM.

landscape/client/configuration.py
silverdrake11: Totally forgot about this. Done

wck0 (approved):

PR #245, opened 2024-05-20
ci: add nightly builds for total-project TICS static analysis

srunde3:
- Add a workflow to run TICS nightly for the entire project.
- Move CI-specific dependencies in new `depends-ci` recipe and move `pre-commit` into `depends-dev` recipe.

There isn't much manual testing that can be done here; feel free to check my fork for successful runs on the `nightly-build-tiobe` branch under the "Actions" tab. (https://github.com/srunde3/landscape-client/actions/runs/9139973764)
>Note that only runs from the `night-build-tiobe` branch are relevant

I needed to add some QA suppression for a couple of older files to get the gatekeeping feature to pass:
- `lib/compat` is for Python 2/3 compatibility, so the Python 2 code branch fails to run under TICS, which attempts to run all code branches under Python 3.
- `jiffies` and `computertags` have style warnings that are critical enough to fail the pipeline, so we will suppress them to keep the pipeline green and informative towards new changes

Some behind-the-scenes work that's referenced in this merge:
- The Tiobe auth token is present as a repo secret (already done in canonical/landscape-client)
- The `tics: disabled` comment at the top of a file is a keyword that we coordinated with the Tiobe server config to ignore checking for that file


Makefile
```diff
@@ -136,6 +143,13 @@ snap:
 	$(SNAPCRAFT)
 .PHONY: snap
 
+# TICS expects coverage info to be in ./coverage/.coverage
+tics-analysis: depends-ci coverage
+	sudo apt install pylint
+	mkdir -p coverage
+	mv .coverage ./coverage/.coverage
```
abehnia, line 150: We may want to consider copying these instead. I know this is CI specific, but in case we may want to run this locally. I also think `clean` should be updated to remove the copied files in case they exist. Add to `gitignore` as well.

landscape/client/monitor/computertags.py
```diff
@@ -15,7 +15,7 @@ class ComputerTags(DataWatcher):
     run_interval = 3600  # Every hour only when data changed
     run_immediately = True
 
-    def __init__(self, args=sys.argv):
```
abehnia, line 18: Regarding these three files, can we try to fix the issues instead of bypassing TICS?

abehnia (changes requested):
Left inline comments.

Makefile
srunde3: Sure.  This is a good reminder to tweak the name a bit, too.  It doesn't really  *run* TiCS analysis, it just prepares the workspace for it.  So I'll name it `prepare-tics-analysis` instead.

landscape/client/monitor/computertags.py
srunde3: the `args=sys.argc` could probably be fixed easily; I'm hesitant to change the other two, though.  The compatibility file necessarily has some Python 2 code, so TICS will complain about that when it runs under Python 3.  `os._exit(0)` has somewhat unique behavior over other system exits, so I don't want to modify that.

In general I prefer to keep the scope of this merge narrow so that there are only CI changes.  IMO another MR would be best to clean up the suppressions with code changes.

srunde3:
FYI I'll need to re-run TICS on my fork to make sure the major version on that action didn't affect anything

landscape/client/monitor/computertags.py
abehnia: Sure, let's keep it CI specific but follow it up with a ticket.

abehnia (changes requested):
Let's add the `coverage` folder to `.gitignore`.

srunde3:
Fixed the .gitignore.  I snuck in a fix to the mutable arg `sys.argv` suppression since that's pretty straightforward and low-risk.  RE `os._exit(0)`; that appears to be a valid use case for `os._exit` (https://docs.python.org/3/library/os.html#os._exit), so I think we might be stuck suppressing that.  I wouldn't plan on making a ticket for it, then.

landscape/client/monitor/computertags.py
srunde3: Snuck in a fix for this so we don't have to suppress it

abehnia (approved):
LGTM. :ship:

PR #250, opened 2024-07-08
chore: update manpages

Perfect5th:
Updated manpages via `make manpages`

Took suggestion from comment on #160 to remove manage-source-list-d entry.


srunde3 (approved):
LGTM

PR #251, opened 2024-07-22
fix: revert fix for (LP: #1827857)

wck0:
The fix was implemented in the `python3-apt` package. The changes in `landscape-client` now cause issues with package reporter on machines with certain locale configurations.


Perfect5th:
Verified that this is effective in `noble`. Is it also effective in other series, did they get the `python-apt` fix?

wck0:
The fix to `python-apt` was added shortly after this workaround was merged into the client codebase. I think it was only a matter of days. Reference comment #4 here https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/1827857

Perfect5th (approved):
Right. Looks like that made it all the way back to focal. LGTM, then.

PR #253, opened 2024-07-29
fix: patch missing Trial addDuration method in py312

Perfect5th:
Twisted Trial has a bug where the `_AdaptedReporter` is missing the `addDuration` method. A warning is emitted for this in Python 3.12. This patch adds a no-op one to eliminate these warnings. It can be removed once this is fixed in Twisted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ PYDOCTOR ?= pydoctor
 TXT2MAN ?= txt2man
 PYTHON ?= python3
 SNAPCRAFT = SNAPCRAFT_BUILD_INFO=1 snapcraft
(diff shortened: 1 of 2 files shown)


wck0 (approved):
I tested via`python3 -m twisted.trial landscape.lib.apt.package.tests.test_facade.AptFacadeTest.test_custom_root_create_required_files` test here, but I still see the warning about `addDuration`, but I suppose that's to be expected.

It is certainly absent when running `make check`

wck0:
Maybe we can also update the README to suggest using `python3 -m landscape.lib.run_tests` to run specific tests.

PR #254, opened 2024-07-29
fix: ensure preferences.d directory exists in created APT dirs

Perfect5th:
Ubuntu 24.04 and higher expect the `preferences.d` directory to exist in any directory being treated as an APT config directory. Creation of this directory was missing in `AptFacade`, which was leading to warnings when running some tests, such as `andscape.lib.apt.package.tests.test_facade.AptFacadeTest.test_custom_root_create_required_files`.


wck0 (approved):
Verified the warning on the specified test is no longer emitted in noble after this change.

:shipit:

PR #255, opened 2024-07-29
fix: add Date and hash fields to tests that create repositories

Perfect5th:
apt-cache expects all package repositories to have the `Date` field and some kind of hash (`MD5Sum`, `SHA1`, `SHA256`) field in their `Release` files. It will emit a warning it either of these are missing, which was polluting our test output. This patch adds these fields in tests to satisfy apt-cache.


wck0 (approved):
:shipit:

PR #256, opened 2024-07-31
fix: reduce computer title retry delay

st3v3nmw:
Got feedback [here](https://forum.snapcraft.io/t/reboot-required-before-landscape-client-device-available-in-backend/41069/11) to reduce the delay between retries. The initial 2 minutes was a little too long.. Now we'll retry after 15s, then 30, 60, etc over a span of ~30 minutes.


Perfect5th (approved):
LGTM. Thanks for adding the 24.04 Ubuntu base to the CI matrix, too.

PR #257, opened 2024-08-01
fix: get ubuntu pro info on snaps

st3v3nmw:
[The Bug Report](https://bugs.launchpad.net/landscape-client/+bug/2074026)

This fix will allow the snap to push Ubuntu Pro information to Landscape by reading the `/var/lib/ubuntu-advantage/status.json` file. We have a `system-files` plug for this file.

One caveat is that the `data_dir` setting (`/var/lib/ubuntu-advantage/`) can be changed in `/etc/ubuntu-advantage/uaclient.conf` which would prevent the snap from reading the status file, as we can't plug into arbitrary folders on the host fs, especially not at runtime.

<details>
<summary>Sample Output (click to expand)</summary>

image

</details>

### DIFF
--- a/landscape/client/__init__.py
+++ b/landscape/client/__init__.py
@@ -9,3 +9,9 @@ GROUP = "root" if IS_SNAP else "landscape"
 DEFAULT_CONFIG = (
     "/etc/landscape-client.conf" if IS_SNAP else "/etc/landscape/client.conf"
 )
+
+UA_DATA_DIR = (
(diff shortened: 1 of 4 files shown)


Perfect5th (approved):
LGTM, thanks!

st3v3nmw:
Thanks @Perfect5th. I've made a request for auto-connection [here](https://forum.snapcraft.io/t/ubuntu-pro-status-system-files-request-for-landscape-client/41279).

PR #261, opened 2024-08-12
build: use dev-lxc to fully configure container

astok:
The configuration of the container for Landscape client would use `dev-lxc` to set up the container, but would then require the user to run a few more manual steps. Here, those additional manual steps are now done automatically. The command should be run as:
```
dev_lxc create --config ./.dev-lxc/config.yaml jammy
```

### DIFF
--- a/.dev-lxc/config.yaml
+++ b/.dev-lxc/config.yaml
@@ -1,0 +1,11 @@
+config:
+  user.user-data: |
+    #cloud-config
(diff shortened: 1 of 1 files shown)


Perfect5th (approved):
LGTM. Results in a container that immediately passes `make check` unit tests after `dev_lxc create --config .dev-lxc/config.yaml jammy`

PR #262, opened 2024-08-19
refactor: use synchronous message exchange for landscape-config

Perfect5th:


Perfect5th:
This should more-or-less be a drop-in refactor, main differences being that registration does not depend on `landscape-client.service` starting, though it is still started after registration succeeds

To test, try `sudo ./scripts/landscape-config`

landscape/lib/compat.py
wck0, line 1: Can we add a task to remove this file?

wck0 (approved):
Manual testing went well.

LGTM

landscape/lib/compat.py
Perfect5th: Will do.

PR #266, opened 2024-09-06
fix: parse livepatch process status check

HJK-X:
The LivePatch plugin runs `canonical-livepatch status --format humane` to get kernel information. 
```
$ canonical-livepatch status --format humane
last check: 54 minutes ago
kernel: 6.8.0-40.40-generic
server check-in: succeeded
kernel state: ✓ kernel series 6.8 is covered by Livepatch
patch state: ✓ no livepatches available for kernel 6.8.0-40.40-generic
tier: updates (Free usage; This machine beta tests new patches.)
machine id: asdfsasdfasdfasdf
```

However, if this command fails, we could not parse that output to still receive other information, and would result in a yaml parser error.
```
$ canonical-livepatch status --format humane
last check: 50 minutes ago
kernel: 6.8.0-41.41-generic
server check-in: failed: livepatch check failed: POST request to "badurl.com/v1/client/asdfsasdfasdfasdf/updates" failed
kernel state: ✓ kernel series 6.8 is covered by Livepatch
patch state: ✓ no livepatches available for kernel 6.8.0-41.41-generic
tier: updates (Free usage; This machine beta tests new patches.)
machine id: asdfsasdfasdfasdf
```
Note that due to additional colons in the "server check-in" entry, yaml.safe_load would not be able to parse this output.

This fix is a custom parser to read the entries in place of yaml.safe_load to handle if the livepatch status command fails.

Manual Test:
1. Start server instance with freshdata
2. Edit client's .conf file to connect to server
3. Register a dev client with the livepatch plugin (enabled by default)
4. [Install](https://linuxhandbook.com/livepatch-ubuntu-server/) and [enable](https://auth.livepatch.canonical.com/) livepatch in the dev environment
5. `sudo canonical-livepatch config remote-server="badurl.com"`
6. `sudo ./scripts/landscape-client`
7. Get JWT token
8. `http GET "http://<server ip>:9091/api/v2/computers/<computer id>"  Authorization:"Bearer $JWT"`
9. Find the livepatch information and check to see that server check-in contains a failure message

Reference:
```
$ JWT=$(curl -X POST "http://<server ip>:8080/api/v2/login" -d '{"email": "john@example.com", "password": "pwd"}' | grep '"token": "[^"]*"' | awk -F'"token": "' '{print $2}' | awk -F'"' '{print $1}')
$ http GET "http://<server ip>:9091/api/v2/computers/<computer id>"  Authorization:"Bearer $JWT" | jq ".livepatch_info"
```

### DIFF
--- a/landscape/client/manager/livepatch.py
+++ b/landscape/client/manager/livepatch.py
@@ -29,15 +29,15 @@ class LivePatch(DataWatcherManager):
 
 
 def _parse_humane(output):
-    entries = output.split("\n")
+    entries = output.splitlines()
     data = {}
 
-    for e in entries:
-        line = e.strip()
-        if len(line) == 0:
+    for entry in entries:
+        line = entry.strip()
+        if line == "":
             continue
-        elif ": " in line:
-            key, value = line.split(": ", 1)
+        elif ":" in line:
+            key, value = line.split(":", 1)
             key = key.strip().strip('"').strip("'")
             value = value.strip().strip('"').strip("'")
             data[key] = value
--- a/landscape/client/manager/livepatch.py
+++ b/landscape/client/manager/livepatch.py
@@ -2,8 +2,6 @@ import json
 import logging
 import subprocess
 import traceback
-import yaml
-
 
 from landscape.client.manager.plugin import DataWatcherManager
 
@@ -22,10 +20,22 @@ class LivePatch(DataWatcherManager):
     run_interval = 1800  # Every 30 min
 
     def get_data(self):
-        json_output = get_livepatch_status('json')
-        readable_output = get_livepatch_status('humane')
-        return json.dumps({'humane': readable_output, 'json': json_output},
-                          sort_keys=True)  # Prevent randomness for cache
+        json_output = get_livepatch_status("json")
+        readable_output = get_livepatch_status("humane")
+        return json.dumps(
+            {"humane": readable_output, "json": json_output}, sort_keys=True
+        )  # Prevent randomness for cache
+
+
+def _parse_humane(cmd_output):
+    entries = cmd_output.split("\n")
+    data = {}
+
+    for e in entries:
+        key, value = e.split(":", 1)
+        data[key] = value.strip()
+
+    return data
 
 
 def get_livepatch_status(format_type):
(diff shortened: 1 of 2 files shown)


landscape/client/manager/livepatch.py
```diff
@@ -22,10 +21,30 @@ class LivePatch(DataWatcherManager):
...
-                          sort_keys=True)  # Prevent randomness for cache
+        json_output = get_livepatch_status("json")
+        readable_output = get_livepatch_status("humane")
+        return json.dumps(
+            {"humane": readable_output, "json": json_output}, sort_keys=True
+        )  # Prevent randomness for cache
+
+
+def _parse_humane(output):
+    entries = output.split("\n")
```
silverdrake11, line 32: splitlines() would be better here as there is some extra logic in there besides the "\n"

landscape/client/manager/livepatch.py
```diff
@@ -22,10 +21,30 @@ class LivePatch(DataWatcherManager):
...
+    data = {}
+
+    for e in entries:
+        line = e.strip()
+        if len(line) == 0:
```
silverdrake11, line 37: len(line) == 0 would always evaluate to empty string '' so it's more clear to `if not line` or to `if line == ''`

landscape/client/manager/livepatch.py
```diff
@@ -22,10 +21,30 @@ class LivePatch(DataWatcherManager):
...
+            continue
+        elif ": " in line:
```
silverdrake11, line 39: It's probably more safe to split on just ":" in case a space would be missing

landscape/client/manager/livepatch.py
```diff
@@ -22,10 +21,30 @@ class LivePatch(DataWatcherManager):
...
+            key, value = line.split(": ", 1)
```
silverdrake11, line 40: See above comment

silverdrake11: You can strip() after to get rid of the space

landscape/client/manager/livepatch.py
```diff
@@ -22,10 +21,30 @@ class LivePatch(DataWatcherManager):
...
+        return json.dumps(
+            {"humane": readable_output, "json": json_output}, sort_keys=True
+        )  # Prevent randomness for cache
+
+
+def _parse_humane(output):
+    entries = output.split("\n")
+    data = {}
+
+    for e in entries:
```
silverdrake11, line 35: Single variable names mess up the python debugger, among other things. Best something like "en" instead of "e" or something else

silverdrake11:
It looks good to me to merge in except for one more request, at this line raise yaml.YAMLError("Input is not a list of key/value pairs"), add the contents of it. So if we see that error show up, we can see exactly why. so like yaml.YAMLError("Input is not a list of key/value pairs" + output)

silverdrake11 (approved):
LGTM

PR #269, opened 2024-09-12
fix: tests pass in a build environment

wck0:
Some recent changes caused the tests to fail in a package building environment where there is no landscape user. These changes allow setting environment variables to configure the user that runs in a build environment.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -26,4 +26,4 @@ override_dh_installsystemd:
 	dh_installsystemd
 
 override_dh_auto_test:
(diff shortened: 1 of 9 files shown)


wck0:
To test, follow our debian build instructions to the point where you can do `pbuilder-dist jammy amd64 build landscape_client*.dsc` and check the build is successful.

srunde3 (approved):
LGTM. Build was successful and tests pass under `pbuilder-dist`.

PR #271, opened 2024-09-17
Add annotations content interface to snapcraft.yaml

mcw-work:
Added a content interface to snapcraft.yaml to allow another snap to access the annotations directory.


Perfect5th (approved):
LGTM

PR #273, opened 2024-10-01
feat: add cloud-init message

abehnia:
Based on `livepatch` and `ubuntupro` this was created as a manager but it can be changed to monitor as well.

### DIFF
--- a/landscape/client/manager/cloudinit.py
+++ b/landscape/client/manager/cloudinit.py
@@ -1,0 +1,66 @@
+import json
+import subprocess
+from typing import Any
+
+from landscape.client.manager.plugin import DataWatcherManager
+
+
+class CloudInit(DataWatcherManager):
+
+    message_type = "cloud-init"
+    message_key = message_type
+    scope = "cloud-init"
+    persist_name = message_type
+    run_immediately = True
+    run_interval = 3600 * 24  # 24h
+
+    def get_data(self) -> str:
+        return json.dumps(get_cloud_init(), sort_keys=True)
+
+
+def get_cloud_init() -> dict[str, Any]:
+    """
+    cloud-init returns all the information the instance has been initialized
+    with, in JSON format. This function takes the the output and parses it
+    into a python dictionary and sticks it in "output" along with error and
+    return code information.
+    """
+
+    data: dict[str, Any] = {}
+    output: dict[str, Any] = {}
+
+    try:
+        completed_process = subprocess.run(
+            ["cloud-init", "query", "-a"],
+            encoding="utf8",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+    except FileNotFoundError as exc:
+        data["return_code"] = -1
+        data["error"] = str(exc)
+        data["output"] = ""
+    except Exception as exc:
+        data["return_code"] = -2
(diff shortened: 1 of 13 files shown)


silverdrake11:
I have not looked at your code but livepatch and ubuntu pro are manager because they need to admin access. "pro status" works on the current user like "ubuntu" in an lxd but does not work with the "landscape" user. To test whether manager is needed, try running the command with the "landscape" user, if you get an error then continue to use manager, if it works fine, then monitor would be best as it uses the least privileges

landscape/client/manager/cloudinit.py
```diff
@@ -0,0 +1,66 @@
...
+        data["return_code"] = -2
+        data["error"] = str(exc)
+        data["output"] = ""
+    else:
+        string_output = completed_process.stdout.strip()
+        try:
+            # INFO: We don't want to parse an empty string.
+            if string_output:
+                json_output = json.loads(string_output)
+                # INFO: Only return relevant information from cloud init.
```
silverdrake11, line 53: This is a good call as there can be sensitive data in the cloud init file

landscape/client/manager/cloudinit.py
```diff
@@ -0,0 +1,66 @@
...
+                output["availability_zone"] = json_output.get(
+                    "availability_zone",
+                    "",
+                ) or json_output.get("availability-zone", "")
```
silverdrake11, line 57: Seems like there can be dashes or underscores in the cloud init file?

landscape/client/manager/cloudinit.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class CloudInit(DataWatcherManager):
+
+    message_type = "cloud-init"
+    message_key = message_type
+    scope = "cloud-init"
+    persist_name = message_type
+    run_immediately = True
+    run_interval = 3600 * 24  # 24h
```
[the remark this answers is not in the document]
silverdrake11: I suppose the cloud init cannot change though? If that's the case then 24 hours is fine.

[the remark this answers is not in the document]
abehnia: Yes, it should be ready-only.

landscape/client/manager/cloudinit.py
abehnia: Yes.

landscape/client/monitor/cloudinit.py
```diff
@@ -0,0 +1,69 @@
...
+    message_key = message_type
+    scope = message_type
+    persist_name = message_type
+    run_immediately = True
+    run_interval = 3600 * 24  # 24h
+
+    def get_data(self) -> str:
+        return json.dumps(get_cloud_init(), sort_keys=True)
+
+    def register(self, monitor):
```
silverdrake11, line 19: You shouldn't need this register. See https://github.com/canonical/landscape-client/pull/130/files and https://github.com/canonical/landscape-client/pull/158

silverdrake11 (changes requested):

landscape/client/monitor/cloudinit.py
abehnia: Went over offline.

silverdrake11 (approved):
LGTM

PR #274, opened 2024-10-01
fix: landscape-config uses ssl_public_key if provided

wck0:
### Current behavior:
Verify that prior to this commit, `landscape-config` will fail if a self-signed cert is required and present in the config.

Try setting `ssl_public_key` in `client.conf` and calling `landscape-config --silent`

The remove `ssl_public_key` from `client.conf` and try calling `landscape-config --silent --ssl-public-key /path/to/your.pem`

In both cases, the connection will fail because of certificate verification failure.

### With this PR:

Repeat the same two calls above and verify the registration message sends successfully.

### DIFF
--- a/landscape/client/configuration.py
+++ b/landscape/client/configuration.py
@@ -3,7 +3,6 @@
 This module, and specifically L{LandscapeSetupScript}, implements the support
 for the C{landscape-config} script.
 """
-
 import getpass
 import io
 import logging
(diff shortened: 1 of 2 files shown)


HJK-X (approved):
tested, LGTM

Perfect5th (approved):
+1 LGTM

PR #280, opened 2024-10-15
fix: add missing python3-yaml to runtime dependencies (LP: #2084586)

Perfect5th:
python3-yaml is required for snap management when parsing snap assertions. Without it, the landscape-client service fails to start. It's often already installed in Ubuntu installations, so sometimes it works without being an explicit dependency.
This change ensures that it is an installed dependency.


Perfect5th:
See https://bugs.launchpad.net/ubuntu/+source/landscape-client/+bug/2084586

silverdrake11 (approved):
LGTM

PR #284, opened 2024-11-07
Fix: insecure id not set during registration on self hosted

david-mclain:
### Verify Bug
- Follow instructions [here](https://ubuntu.com/landscape/docs/install-landscape-client#heading--install-landscape-client-from-a-PPA) to install client and use `ppa:landscape/self-hosted-beta`
- Connect client to ppa beta version of self hosted server following instructions [here](https://ubuntu.com/landscape/docs/quickstart-deployment)
- After connecting client wait a few minutes and client should go offline as it is not pinging the server

### Verify fix
- Use this version of client
- Connect client to self hosted server
- Verify ping is working and connection doesn't go offline

### DIFF
--- a/landscape/client/tests/test_configuration.py
+++ b/landscape/client/tests/test_configuration.py
@@ -26,6 +26,7 @@ from landscape.client.configuration import set_secure_id
 from landscape.client.configuration import setup
 from landscape.client.configuration import show_help
 from landscape.client.configuration import store_public_key_data
+from landscape.client.registration import RegistrationInfo
 from landscape.client.serviceconfig import ServiceConfigException
 from landscape.client.tests.helpers import LandscapeTest
 from landscape.lib.compat import ConfigParser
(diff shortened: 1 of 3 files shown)


landscape/client/tests/test_configuration.py
```diff
@@ -1496,6 +1497,44 @@ def test_register_silent(
...
+        self.assertRaises(
+            SystemExit,
+            main,
+            ["--silent", "-c", self.make_working_config()],
+            print=noop_print,
+        )
+
+        mock_setup.assert_called_once()
+        mock_input.assert_not_called()
+        mock_set_secure_id.assert_called_once_with(mock.ANY, mock.ANY, 10)
```
Perfect5th, line 1536: Probably safest to assert that it's called with `mock.ANY, "fake-secure-id", 10`.

Perfect5th (changes requested):
One minor comment on test. Otherwise looks good.

Perfect5th (approved):

PR #285, opened 2024-11-14
Fix: monitor-only not being set properly when false/no passed in config

david-mclain:
### Verify Bug
- Run new testcases provided (current branch) checking for when false parameter is sent in using old watchdog file
- Test cases will fail as Manager daemon is not spawned
### Verify Fix
- Run same testcases as above using new watchdog
- See that Manager daemon is spawned in cases where false is sent as parameter

### DIFF
--- a/landscape/client/tests/test_deployment.py
+++ b/landscape/client/tests/test_deployment.py
@@ -4,6 +4,7 @@ from unittest import TestCase
 
 from landscape.client.deployment import BaseConfiguration
 from landscape.client.deployment import Configuration
+from landscape.client.deployment import convert_arg_to_bool
 from landscape.client.deployment import generate_computer_title
 from landscape.client.deployment import get_versioned_persist
 from landscape.client.deployment import init_logging
@@ -468,7 +469,7 @@ class GenerateComputerTitleTest(TestCase):
         )
         self.assertIsNone(title)
         mock_debug.assert_called_once_with(
-            "No serial assertion in snap info {}, waiting..."
+            "No serial assertion in snap info {}, waiting...",
         )
 
     @mock.patch("landscape.client.deployment.debug")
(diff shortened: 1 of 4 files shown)


landscape/client/watchdog.py
```diff
@@ -535,6 +537,22 @@ def get_enabled_daemons(self):
             daemons.append(Manager)
         return daemons
 
+    def convert_arg_to_bool(self, value: str) -> bool:
```
srunde3, line 540: Can we move this into `landscape/client/deployment.py` and make it a function? I don't think it really needs to be a method on a particular configuration type. It's more generic than that.

We should also add unit tests specifically for this function.

If you'd like to mimic Python's configparser here, you can use these values:

```python
# Case-insensitive values to use when parsing booleans.
TRUTHY_VALUES = {"true", "yes", "y", "1", "on"}
FALSY_VALUES = {"false", "no", "n", "0", "off"}
```

We had to implement something similar for server in `canonical/lib/config.py`. You can use the unit tests there for reference on these unit tests, too.

srunde3 (changes requested):
Left note inline for a refactor + some unit tests.

landscape/client/tests/test_deployment.py
```diff
@@ -601,3 +602,33 @@ def test_generate_computer_title_with_date(
...
+        for t in TRUTHY_VALUES:
+            val = convert_arg_to_bool(t)
+        self.assertTrue(val)
+
+    def test_false_values(self):
+        FALSY_VALUES = {"false", "no", "n", "0", "off", "FALSE", "No"}
+        val = False
+        for f in FALSY_VALUES:
+            val = convert_arg_to_bool(f)
+        self.assertFalse(val)
```
srunde3, line 622: We'll want this in the loop - it's OK if the test fails after the first failing instance. We don't have to worry about subtests.

You also won't need to initialize `val`, then.

landscape/client/tests/test_deployment.py
```diff
@@ -601,3 +602,33 @@ def test_generate_computer_title_with_date(
...
+    def test_invalid_values(self, logging):
+        INVALID_VALUES = {"invalid", "truthy", "2", "exit"}
+        val = False
+        for i in INVALID_VALUES:
+            val = convert_arg_to_bool(i)
+            logging.assert_called_with(
+                "Error. Invalid boolean provided in config or parameters. "
+                + "Defaulting to False.",
+            )
+        self.assertFalse(val)
```
srunde3, line 634: same here

landscape/client/tests/test_deployment.py
```diff
@@ -601,3 +602,33 @@ def test_generate_computer_title_with_date(
...
+
+class ArgConversionTest(LandscapeTest):
+    """Tests for `convert_arg_to_bool` function"""
+
+    def test_true_values(self):
+        TRUTHY_VALUES = {"true", "yes", "y", "1", "on", "TRUE", "Yes"}
+        val = True
+        for t in TRUTHY_VALUES:
+            val = convert_arg_to_bool(t)
+        self.assertTrue(val)
```
srunde3, line 615: same here

srunde3 (changes requested):
small notes inline for the tests.

srunde3 (approved):
LGTM :ship: :hippopotamus: 

Manual testing went well - I ran `scripts/landscape-client` with various `--monitor-only` values to confirm that Manager plugins only run when it's `false`.

PR #286, opened 2024-11-22
Fix: SSL Certificates not properly surfaced

david-mclain:
### Verify Bug
- Download landscape client and place self signed server certificate in an unreadable directory by landscape
- Try to connect client to server using the following command
`landscape-config --account standalone --url https://<server-ip>/message-system --ping-url https://server-ip/ping -k ssl-cert-location `
- Code should fail and provide no information as to why it fails
### Verify Fix
- Repeat above steps but instead using this version
- Verify that warning message is provided by logs

### DIFF
--- a/landscape/lib/fetch.py
+++ b/landscape/lib/fetch.py
@@ -2,6 +2,7 @@ import io
 import os
 import sys
 from argparse import ArgumentParser
+from logging import warning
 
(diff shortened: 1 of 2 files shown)


srunde3 (approved):
LGTM - I tested using a file path that didn't exist, and that also logged a warning message (as expected). :rocket:

PR #287, opened 2024-11-22
fix: TiOBE TiCS workflow

wck0:
You can view a successful run here: https://github.com/wck0/landscape-client/actions/runs/11977115496


HJK-X (approved):
LGTM

PR #288, opened 2024-11-25
fix: fixed issues where --is-registered would print incorrect outputs

david-mclain:
### Verify Bug
- Register a new client to a server, without accepting the new computer on the server side yet
- Run `landscape-config --is-registered`, verify output says True
### Verify Fix
**_NOTE:_**  In order to not disrupt any workflows we added new command line arguments rather than editing existing ones
- Repeat registration without accepting step
- Run `landscape-config --actively-registered` and verify output is false
- Run `landscape-config --registration-sent` and verify output is true
- Accept the new computer on server end, and rerun `landscape-config --actively-registered`, verify output is True

### DIFF
--- a/landscape/client/configuration.py
+++ b/landscape/client/configuration.py
@@ -788,7 +788,8 @@ def registration_sent(config):
     """
     Return whether the client has sent a registration request to the server.
     For now does same thing as is_registered as to make function name more
-    clear with what is performed.
+    clear with what is performed. This is the legacy behaviour of
+    --is-registered and the name will be changed in a future release.
     """
     persist_filename = os.path.join(
         config.data_path,
@@ -810,17 +811,6 @@ def actively_registered(config):
     if accepted_types is not None:
         return len(accepted_types) > 1 and "register" not in accepted_types
     return False
-
-
-def is_registered(config):
-    """Return whether the client is already registered."""
-    persist_filename = os.path.join(
-        config.data_path,
-        f"{BrokerService.service_name}.bpickle",
-    )
-    persist = Persist(filename=persist_filename, user=USER, group=GROUP)
-    identity = Identity(config, persist)
-    return bool(identity.secure_id)
 
 
 def registration_info_text(config, registration_status):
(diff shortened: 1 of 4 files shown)


man/landscape-config.txt
```diff
@@ -83,6 +83,12 @@ OPTIONS
   --is-registered          Exit with code 0 (success) if client
                            is registered else returns 5. Display
```
Perfect5th, line 84: should probably update this to match the "real" behaviour:
```
Exit with code 0 (success) if client
has sent registration request else returns 5.
Display registration sent info.
(NOTE: use --actively-registered to detect if
registration has been accepted server-side)
```
something like that.

Perfect5th (changes requested):

Perfect5th (approved):

landscape/client/configuration.py
```diff
@@ -871,6 +903,17 @@ def main(args, print=print):
         else:
             sys.exit(EXIT_NOT_REGISTERED)
 
+    if config.actively_registered:
+        currently_registered = actively_registered(config)
+
+        info_text = registration_info_text(config, currently_registered)
+        print(info_text)
+
+        if registration_status:
```
Perfect5th, line 912: this needs to be changed to `if currently_registered`. In the case where `config.is_registered` and `config.registration_sent` are both `False` and `config.actively_registered` is `True`, it is undefined, though the script still prints the status:
```
$ sudo ./scripts/landscape-config --actively-registered

Registered:    False
Config Path:   /home/ubuntu/landscape-client/landscape-client.conf
Data Path      /tmp/landscape/
  File "/home/ubuntu/landscape-client/./scripts/landscape-config", line 16, in <module>
    main(sys.argv[1:])
  File "/home/ubuntu/landscape-client/./landscape/client/configuration.py", line 912, in main
    if registration_status:
UnboundLocalError: local variable 'registration_status' referenced before assignment
```

Perfect5th (changes requested):

Perfect5th (approved):

PR #291, opened 2024-11-27
fix: Updated motd

david-mclain:
edited string for motd to point to right link

### DIFF
--- a/landscape/sysinfo/landscapelink.py
+++ b/landscape/sysinfo/landscapelink.py
@@ -7,7 +7,7 @@ class LandscapeLink:
 
     def run(self):
         self._sysinfo.add_footnote(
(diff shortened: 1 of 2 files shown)


landscape/sysinfo/landscapelink.py
```diff
@@ -7,7 +7,7 @@ def register(self, sysinfo):
 
     def run(self):
         self._sysinfo.add_footnote(
-            "Graph this data and manage this system at:\n"
```
Perfect5th, line 10: definitely opinion-based, but I kind of like having the `\n` to make the URL stand out a little more (and avoid possibly wrapping it on narrowing consoles).

Perfect5th (approved):
Approved with one optional suggestion.

PR #295, opened 2025-01-09
Update snapcraft.yaml

mcw-work:
Fix: removed snapd-control-managed from apps stanza. This still works but means that the snap declaration on the store does not need changing - which seems to break existing installations.


mcw-work:
@Perfect5th  ^^ linting has failed on a fille I haven't touched.....  and looks to have not been touched for 2 years. I would try it again, but I don't have the power.

Perfect5th (approved):

PR #299, opened 2025-01-27
Add USG plugin

Perfect5th:


Perfect5th:
There's no way to conveniently manually test this plugin yet - it's not active and there's no server-side message. Those will be defined in the near future.

Perfect5th:
I've got it as a follow-up action to refactor the attachment-handling code to make it more generally useful and interface-y.

landscape/client/manager/tests/test_usgmanager.py
```diff
@@ -0,0 +1,358 @@
...
+        set.
+        """
+        which = mock.patch("shutil.which").start()
+        path_touch = mock.patch(MODULE + ".Path.touch").start()
+        send_message = mock.patch.object(
+            self.plugin.registry.broker,
+            "send_message",
+            new=mock.AsyncMock(),
+        ).start()
+        pkg_file = os.path.join(self.tempdir, "reboot-required.pkgs")
```
wck0, line 160: nit

landscape/client/manager/tests/test_scriptexecution.py
```diff
@@ -302,99 +306,6 @@ def check(result):
         result.addCallback(check)
         return result
 
-    def test_run_with_attachment_ids(self):
```
wck0, line 305: These tests were basically moved to the tests for the new `attachments` module, correct?

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def _get_last_audit_results(self) -> str | None:
+        """Returns the file path of the most recently produced audit report. If
+        no audit reports exist, returns `None`.
+        """
+        audit_files = glob.glob(USG_AUDIT_RESULTS_GLOB)
+        if not audit_files:
+            return None
+
+        last_audit_results = max(audit_files, key=os.path.getctime)
```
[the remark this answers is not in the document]
Perfect5th: 1. one per audit, which will be configurable via the scheduling of the Landscape USG profile. Could be many.
2. not that I'm aware of. I don't think they _should_ be cleaned up either.
3. seems incredibly unlikely… if it does happen than an error will be reported as a result of the activity, though.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,245 @@
...
+    async def _save_attachment(
+        self,
+        attachment: Tuple[str, int] | None,
+    ) -> str | None:
+        """Downloads `attachment` from Landscape Server and saves it in a
+        tempfile.
+
+        :param attachment: A tuple of filename and attachment ID.
+
+        :returns: The tempfile's path or `None` if no attachment exists.
```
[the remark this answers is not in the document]
Perfect5th: you're right - I forgot to change this when I moved from using a tempfile to using a regular file.
We do delete the file, in the `finally` in `_usg_operation`.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,245 @@
...
+        with open(last_result, "rb") as audit_results:
+            message = {"type": "usg-audit", "report": audit_results.read()}
+
+        return await self.registry.broker.send_message(
+            message,
+            self._session_id,
+            True,
+        )
+
+    def _set_reboot_required(self) -> None:
```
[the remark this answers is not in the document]
Perfect5th: That's right - the idea is that the reboot is required by `usg`, and that gets reported through our usual reboot required monitor mechanism.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,245 @@
...
+        path.touch(exist_ok=True)
+
+        if os.path.exists(REBOOT_REQUIRED_PKG_FILE):
+            with open(REBOOT_REQUIRED_PKG_FILE, "r") as reboot_required_pkg:
+                if "usg" in reboot_required_pkg.read():
+                    # it's already in there, no need to add it.
+                    return
+
+        with open(REBOOT_REQUIRED_PKG_FILE, "a") as reboot_required_pkg:
+            reboot_required_pkg.write("usg\n")
```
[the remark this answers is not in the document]
Perfect5th: If they don't then they are violating the Debian manual. Still, it happens enough.

landscape/client/manager/tests/test_scriptexecution.py
Perfect5th: right - the nitty-gritty of download attachments doesn't belong to `scriptexecution` any more.

Perfect5th:
I'm adding the `USG_AUDIT` message type to server_bound message schemas. It's pretty basic, anyways.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,253 @@
...
+        system should reboot due to USG.
+        """
+        path = Path(REBOOT_REQUIRED_FILE)
+        path.touch(exist_ok=True)
+
+        if os.path.exists(REBOOT_REQUIRED_PKGS_FILE):
+            with open(REBOOT_REQUIRED_PKGS_FILE, "r") as reboot_required_pkg:
+                if "usg" in reboot_required_pkg.read():
+                    # it's already in there, no need to add it.
+                    return
```
wck0, line 187: Maybe some package named `rebusgame` needs a reboot? Should we check line-by-line instead?

wck0: Take or leave this: the string `"usg"` might maybe should be a constant.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,253 @@
...
+from landscape.client.attachments import save_attachments
+from landscape.client.manager.plugin import FAILED
+from landscape.client.manager.plugin import ManagerPlugin
+from landscape.client.manager.plugin import SUCCEEDED
+from landscape.client.manager.scriptexecution import (
+    ProcessAccumulationProtocol,
+)
+from landscape.client.manager.scriptexecution import ProcessFailedError
+
+REBOOT_REQUIRED_FILE = "/var/run/reboot-required"
```
wck0, line 23: This constant already exists in `landscape/client/monitor/rebootrequired.py` but it's probably OK not to reorganize constants in this PR.

landscape/client/manager/usgmanager.py
```diff
@@ -0,0 +1,253 @@
...
+            self._session_id,
+            True,
+        )
+
+    async def _save_attachment(
+        self,
+        attachment: Union[Tuple[str, int], None],
+    ) -> Union[str, None]:
+        """Downloads `attachment` from Landscape Server and saves it in a
+        tempfile.
```
wck0, line 132: Did you still forget to change this?

wck0 (changes requested):
Sorry, a few more comment/asks/suggestions

landscape/client/manager/usgmanager.py
Perfect5th: Yes, I have another follow-up action to consolidate things related to reboot-required.

landscape/client/manager/usgmanager.py
Perfect5th: I thought I did - whoops. Will fix.

landscape/client/manager/usgmanager.py
Perfect5th: Good points, both.

wck0 (approved):
LGTM
:ship: :hippopotamus:

PR #306, opened 2025-03-04
release 24.12+git6418

HJK-X:


wck0 (dismissed):
:shipit:

landscape/__init__.py
```diff
@@ -1,5 +1,5 @@
 DEBIAN_REVISION = ""
```
wck0, line 1: I think we need to restore the `-0landscape0` value here.

wck0 (approved):
:ok:

PR #310, opened 2025-03-26
fix: remove deprecated use of `unittest.makeSuite` to get tests passing on >=py312

Perfect5th:
You can validate this fix pretty easily:

1. create a plucky LXD container:
```bash
dev_lxc create plucky --config ./.dev-lxc/config.yaml
```
3. run the tests:
```bash
make check
```

Without this change, they will fail with:
```
builtins.AttributeError: module 'unittest' has no attribute 'makeSuite'
```

With this change, they should all pass (or skip).


srunde3 (approved):
LGTM. Tests run without issue.

PR #311, opened 2025-03-27
ci: check if CLA is signed on each PR

wck0:
Compare with, e.g., this successful run of the same action in a different repo: https://github.com/canonical/landscape-documentation/actions/runs/14109971121/job/39525837115

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -1,0 +1,13 @@
+name: Check if CLA is signed
+on:
+    pull_request:
+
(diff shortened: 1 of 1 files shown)


srunde3 (changes requested):
Seems like `accept-existing-contributors` is not a valid input and gets ignored - we'd might as well remove it. There's a warning during the run: https://github.com/canonical/landscape-documentation/actions/runs/14109971121

srunde3 (approved):
LGTM

PR #312, opened 2025-04-07
fix: make config tests insensitive to http(s) proxy settings

Perfect5th:
To test:

  1. install an HTTP proxy: `sudo apt install squid`
  2. export environment variables: `export http_proxy=http://localhost:3128 https_proxy=http://localhost:3128`
  3. run the tests, can just be the ones that break: `python3 -m twisted.trial landscape.client.tests.test_configuration`

Without the change, 6 tests will fail. With it, they should all pass.


silverdrake11 (approved):
LGTM

PR #313, opened 2025-04-09
fix: add run-id and operation-id fields to usg-audit message

Perfect5th:
Fairly minor patch to add the `run-id` and `operation-id` fields to the `usg-audit` message type, as they are required for landscape-server to properly associate the results with their origin.


HJK-X (approved):
LGTM

PR #317, opened 2025-04-22
chore: add security policy

wck0:


alex-cendejas (approved):
LGTM, checked links as well, working as expected. Approved.

In the future we may want to enhance this in order to provide more details as found in the SECURITY.md file for LXD, for example. But for now this is OK.

PR #319, opened 2025-04-30
fix: .list and .sources files are restored when a repository profile is disassociated from a noble (or later) instance

jansdhillon:
https://bugs.launchpad.net/landscape-client/+bug/2087852

---

The reason this is only "for noble" is because it uses DEB822-style sources in `/etc/sources.list.d` instead of `sources.list` like jammy and previous. 

## Manual testing

Register a Landscape Client noble instance to Landscape Server.

In the client's `/etc/apt/sources.list.d`:

```bash
touch hello.list hello.sources
```

Create a repository profile, and associate it with the client. The easiest way is through the UI. Wait for the activity to run and in the client's `sources.list.d`, you should see:

```
hello.list.save 
hello.sources.save 
ubuntu.sources.save 
landscape-<profile-name>.list
```

Now, delete the repository profile or disassociate it with the instance, and wait for the activity to run. Now, check on the client's `sources.list.d` You should see that the sources have been restored, and the profile has been deleted:

```
hello.list
hello.sources 
ubuntu.sources
```

Also, it is worth testing the behavior when multiple repository profiles are applied. The most-recently-applied one will take precedent over any others, and when all profiles are removed/disassociated, all `landscape-*.list` profile source files in `/etc/apt/sources.list.d` will be removed and the original sources are restored.

### DIFF
--- a/landscape/client/manager/aptsources.py
+++ b/landscape/client/manager/aptsources.py
@@ -114,6 +114,20 @@ class AptSources(ManagerPlugin):
 
         saved_sources = f"{self.SOURCES_LIST}.save"
 
+        manage_sources_list_d = getattr(
+            self.registry.config,
+            "manage_sources_list_d",
+            True,
+        )
+
+        if manage_sources_list_d not in FALSE_VALUES:
+            for pattern in self.SOURCES_LIST_D_FILE_PATTERNS:
+                filenames = glob.glob(
+                    os.path.join(self.SOURCES_LIST_D, pattern)
+                )
+                for filename in filenames:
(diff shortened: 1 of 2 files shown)


Perfect5th (commented):
Code looks okay to me. Doing some manual testing now.

Perfect5th:
For what's it's worth, I don't think this is just "for noble and later", except if you consider the default case. Earlier series _may_ be able to use DEB822 sources, they just don't do that by default.

Perfect5th (approved):
Manual testing went great. +1 LGTM!

PR #323, opened 2025-04-30
fix: usgmanager should accept only the profile or the tailoring file

HJK-X:
Ensure that usg is enabled:

```sh
sudo pro attach
sudo apt update
sudo apt install ubuntu-advantage-tools
sudo pro enable usg
sudo apt install usg
```

Generate a tailoring file (will be used to generate the security profile in LS)
```sh
sudo usg generate-tailoring disa_stig .
```

connect LC to server and make sure the UsgManager plugin is enabled.
```
include_manager_plugins = UsgManager
```
execute a security profile on server

### DIFF
--- a/landscape/client/manager/usgmanager.py
+++ b/landscape/client/manager/usgmanager.py
@@ -1,4 +1,5 @@
 import glob
+import logging
 import os
 import shutil
 from pathlib import Path
@@ -65,6 +66,7 @@ class UsgManager(ManagerPlugin):
 
         :message: A message of type "usg".
         """
+        logging.info(str(message))
         opid = message["operation-id"]
         runid = message["run-id"]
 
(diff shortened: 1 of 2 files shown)


landscape/client/manager/usgmanager.py
```diff
@@ -202,7 +202,7 @@ def _set_reboot_required(self) -> None:
     def _spawn_usg(
         self,
         action: str,
-        profile: str,
+        profile: str | None,
```
Perfect5th, line 205: Just a heads up: this should be okay unless we need to port to focal, in which case we'll need to use `Optional`.

Perfect5th (commented):
Code review LGTM with one note for the future.

Moving to manual tests.

Perfect5th (approved):

PR #324, opened 2025-05-02
build: release 25.04~rc2

abehnia:


debian/changelog
```diff
@@ -1,3 +1,12 @@
+landscape-client (25.04~rc1-0landscape1) jammy; urgency=medium
+
+  * fix: ensure usgmanager runs in deferred
+  * fix: usgmanager accepts either (exclusive) of the profile or a file
+  * fix: .list and .sources files are restored when a repo profile is
+  disassociated from a noble (or newer) instance
```
wck0, line 6: Is it OK to have a line break here in a debian changelog?

Perfect5th: It's ok, but I usually prefer to align the following line after the `* `:
```
  * fix: .list and .sources files are restored when a repo profile is
    disassociated from a noble (or newer) instance
```

abehnia: changed the alignment

jansdhillon (approved):
🚀

PR #325, opened 2025-05-05
build: release 25.04 for jammy

Perfect5th:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+landscape-client (25.04-0landscape0) jammy; urgency=medium
+
+  * build: 25.04
```
wck0, line 3: `build: release 25.04 for jammy`

wck0: Or just `release 25.04`

Perfect5th: changed to the second suggestion.

wck0 (approved):

PR #326, opened 2025-05-06
fix: invalid plugins do not crash landscape-sysinfo (LP: #1754002)

wck0:
LP bug: https://bugs.launchpad.net/landscape-client/+bug/1754002

You can reproduce this on the main branch in dev using the `scripts/landscape-sysinfo` script and passing invalid parameters. You'll see a TraceBack spewed out.

With this change, passing invalid parameters exits cleanly via the argparser and give some useful feedback instead.


Perfect5th (commented):
Approach seems sound to me. Will complete review once out of draft.

Perfect5th (approved):
:ship:

PR #328, opened 2025-05-13
build: release 25.04+git6435 for jammy

Perfect5th:


debian/changelog
```diff
@@ -1,3 +1,20 @@
...
+  * feat: add usgmanager plugin to example.conf
+  * feat: configurably exclude package sources
+  * fix: remove deprecated use of unittest.makeSuite to get tests passing on
+    >=py312
+  * fix: make config tests insensitive to http(s) proxy settings
+  * fix: add run-id and operation-id fields to usg-audit message
+  * fix: ensure usgmanager plugin runs in deferred
+  * fix: usgmanager should accept only the profile or the tailoring file
+  * fix: .list and .sources files are restored when a repository profile is
+    disassociated from a noble (or later) instance (LP: #2087852)
```
jansdhillon, line 13:
```suggestion
    disassociated from an instance (LP: #2087852)
```

I know this was one of my commits but I think we can leave out the noble part since, as you said, a jammy or lower instance could use deb822-style sources.

jansdhillon (approved):
LGTM

debian/changelog
Perfect5th: The fact that it's your commit makes you more qualified to amend it, not less. Accepting this change.

Perfect5th:
@wck0 is it, isn't it?

jansdhillon (approved):

PR #329, opened 2025-05-27
feat: log pending message count

jansdhillon:
https://warthogs.atlassian.net/browse/LNDENG-2247

---

## Manual testing

Connect a client running this branch to Landscape Server and wait for a couple of minutes. Then, in the Client instance, check for the new log entry in `broker.log`:

```
#root@landscape-client-jammy:/tmp/landscape
cat broker.log | grep "Pending messages"
...
2025-05-27 19:35:07,203 INFO     [MainThread] Pending messages remaining after the last exchange: 1
```


HJK-X (approved):
LGTM

PR #333, opened 2025-06-06
refactor: remove oboslete client-uptime message schema

wck0:
`landscape-client` does not send this message and `landscape-server` does not have a handler for it. In fact, `landscape-server` dropped the corresponding table from the resource database in patch 29.


Perfect5th (approved):
>`# just for backwards compatibility`

Yep, backwards compatibility more than achieved for long enough, I think.

PR #334, opened 2025-06-06
refactor: remove obsolete computer-uptime message schema

wck0:
The associated client plugin has not been enabled for quite some time. See, e.g., https://bugs.launchpad.net/landscape/+bug/911324

Server does not have a handler for this message type, and the corresponding table in the resource database was dropped in patch 88.


Perfect5th (dismissed):
:skull:

Perfect5th (approved):

PR #335, opened 2025-06-06
refactor: remove obsolete hardware-inventory message schema

wck0:
`landscape-client` does not send this message and `landscape-server` does not have a handler for it.

See https://bugs.launchpad.net/landscape/+bug/900655


HJK-X (approved):
:skull:

PR #340, opened 2025-07-02
ci: remove custom CLA check action

wck0:
We appear to now be automatically covered by a check managed by Canonical. Our custom check is therefore redundant as there is no significant difference between the two.
See https://github.com/canonical/landscape-client/actions/workflows/required/canonical/cla-workflow-hosting/.github/workflows/cla.yml


HJK-X (approved):
:boat: :potato:

PR #341, opened 2025-07-11
feat: add machine-id to registration and computer info

wck0:
# Manual Testing Instructions

Enable debug logging in `client.conf` and point it to your Landscape.
Register with Landscape using the dev scripts.
Check the `machine-id` is sent upon registration and then later with the first `computer-info`.
Check the values sent in those two messages match each other.
Check the value differs from your dev machine's real `machine-id` in `/etc/machine-id`


david-mclain (approved):
LGTM :ship: :it:

PR #344, opened 2025-07-24
feat: add --script-tempdir configuration

srunde3:
## Context

Adding a configuration option to set the working directory for script execution improves the CIS hardening experience. There is a "noexec on /tmp" rule that Landscape typically violates by requiring executables (scripts) in `/tmp`.

### Client

Include the following in your `client.conf`:

```ini
script_tempdir = /var/custom-tempdir
manager_plugins = ScriptExecution
script_users = landscape,ALL
```

Set up your custom dir for script execution by the `landscape` user:

```sh
sudo mkdir -p /var/custom-tempdir
sudo chown landscape:landscape /var/custom-tempdir
sudo chmod 700 /var/custom-tempdir
```

Run the client:

```sh
sudo scripts/landscape-client -c landscape-client.conf
```

### Optional: verify file creation in the expected directory

I used `inotify` to verify that the scripts and attachments appeared in my temp directory:

```sh
sudo apt install inotify-tools
sudo inotifywait -m -e create --format '%w%f' /var/custom-tempdir

# Verify that /tmp is not being used by Landscape for scripts
inotifywait -m -e create --format '%w%f' /tmp
```

### Server

Run server (any server) and register your client. Configure a script that uses an attachment. A simple test is to `cat $LANDSCAPE_ATTACHMENTS/<my attachment>` from your script.

Run the script as the `landscape` user. It should succeed. You should be able to see your attachment in the output.

### Extra credit

Repeat, but without the configuration for the tempdirs. Landscape should again use `/tmp`.

### DIFF
--- a/landscape/client/deployment.py
+++ b/landscape/client/deployment.py
@@ -177,7 +177,8 @@ class Configuration(BaseConfiguration):
             "--script-tempdir",
             default=None,
             type=str,
-            help="The working directory to use for script executions.",
+            help="The working directory to use for script executions. "
+            "Must have read, write, and exec privileges for any script users.",
         )
 
         # Hidden options, used for load-testing to run in-process clones
--- a/landscape/client/deployment.py
+++ b/landscape/client/deployment.py
@@ -173,6 +173,19 @@ class Configuration(BaseConfiguration):
             type=int,
             help="The interval between snap monitor runs (default 1800).",
         )
+        parser.add_argument(
+            "--script-tempdir",
+            default=None,
+            type=str,
+            help="The working directory to use for script executions.",
+        )
+        parser.add_argument(
+            "--script-attachment-tempdir",
+            default=None,
+            type=str,
+            help="The working directory to use for script execution "
+            "attachments.",
+        )
 
         # Hidden options, used for load-testing to run in-process clones
         parser.add_argument("--clones", default=0, type=int, help=SUPPRESS)
--- a/landscape/client/deployment.py
+++ b/landscape/client/deployment.py
@@ -179,13 +179,6 @@ class Configuration(BaseConfiguration):
             type=str,
             help="The working directory to use for script executions.",
         )
-        parser.add_argument(
-            "--script-attachment-tempdir",
-            default=None,
-            type=str,
-            help="The working directory to use for script execution "
-            "attachments.",
-        )
 
         # Hidden options, used for load-testing to run in-process clones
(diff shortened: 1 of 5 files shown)


wck0:
Manual testing went fine and I earned the extra credit.

Checking the code changes now...

landscape/client/deployment.py
```diff
@@ -173,6 +173,12 @@ def make_parser(self):
             type=int,
             help="The interval between snap monitor runs (default 1800).",
         )
+        parser.add_argument(
+            "--script-tempdir",
+            default=None,
```
wck0, line 178: I think the default for `default` is `None`, but it's also OK to be explicit about that.

landscape/client/deployment.py
```diff
@@ -173,6 +173,12 @@ def make_parser(self):
...
+            type=str,
+            help="The working directory to use for script executions.",
```
wck0, line 180: Should we inform the user here what the ownership and permissions should be? Or at least allude to them?

wck0 (changes requested):
Some questions in the diff.

landscape/client/manager/scriptexecution.py
```diff
@@ -433,16 +437,17 @@ class ScriptExecution(ManagerPlugin):
...
 
-    def __init__(self):
+    def register(self, client: "BrokerClient"):
         from landscape.client.manager.customgraph import CustomGraphPlugin
 
-        self._script_execution = ScriptExecutionPlugin()
+        super().register(client)
+        self._script_execution = ScriptExecutionPlugin(
+            script_tempdir=self.manager.config.script_tempdir,
+        )
```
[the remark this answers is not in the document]
srunde3: The plugin doesn't actually receive a `config` until `register()` is called on it, so these couldn't be instantiated.

```python
# landscape/client/broker/client.py::BrokerClient

    def add(self, plugin):
        """Add a plugin.

        The plugin's C{register} method will be called with this broker client
        as its argument.

        If the plugin has a C{plugin_name} attribute, it will be possible to
        look up the plugin later with L{get_plugin}.
        """
        info("Registering plugin %s.", format_object(plugin))
        self._plugins.append(plugin)
        if hasattr(plugin, "plugin_name"):
            self._plugin_names[plugin.plugin_name] = plugin
        plugin.register(self)   # <-- here
```

It's possible to get around this by not needing the config in `__init__` of the plugin, but I prefer that the plugins receive their relevant configuration in the `__init__` instead of referencing it throughout the methods.

PR #345, opened 2025-07-25
refactor: change ubuntu pro info to use `uaclient.status` instead of subprocess

david-mclain:
Test cases should cover testing

### DIFF
--- a/landscape/client/manager/ubuntuproinfo.py
+++ b/landscape/client/manager/ubuntuproinfo.py
@@ -33,10 +33,7 @@ class UbuntuProInfo(DataWatcherManager):
                           sort_keys=True)
 
 
-q = Queue()
-
-
-def uastatus():
+def uastatus(q):
     config = UAConfig()
     pro_info = status(config)
     q.put(pro_info)
@@ -104,7 +101,8 @@ def get_ubuntu_pro_info() -> dict:
 
         # The status file has more information than `pro status`
     else:
-        p = Process(target=uastatus)
+        q = Queue()
+        p = Process(target=uastatus, args=(q,))
         p.start()
         p.join()
 
--- a/landscape/client/manager/ubuntuproinfo.py
+++ b/landscape/client/manager/ubuntuproinfo.py
@@ -1,5 +1,4 @@
 import json
-import subprocess
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
@@ -8,6 +7,9 @@ from landscape.client import IS_CORE
 from landscape.client import IS_SNAP
 from landscape.client.manager.plugin import DataWatcherManager
 from landscape.client import UA_DATA_DIR
+
+from uaclient.status import status
+from uaclient.config import UAConfig
 
 
 class UbuntuProInfo(DataWatcherManager):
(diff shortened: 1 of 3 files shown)


Perfect5th:
Calling `uaclient.status.status()` seems to mess with our apt facade and a number of other things. This breaks the tests but also might subtly break things at runtime, which I really don't want.

I suggest we isolate its effects by using a process. It increases overhead, but we were already using a subprocess so I don't think it'll change things much versus what we had before:

simple example

```python
from multiprocessing import Process, Queue

q = Queue()

def uastatus():
    config = UAConfig()
    pro_status = status(config)
    q.put(pro_status)

p = Process(target=uastatus)
p.start()
p.join()

pro_status = q.get()
```

Perfect5th:
Oh and the new status has `datetime` types in it, any time it gets JSON-serialized, you should `json.dumps(ubuntu_pro_info, default=str)` or something like that.

landscape/client/manager/ubuntuproinfo.py
```diff
@@ -65,44 +103,14 @@ def get_ubuntu_pro_info() -> dict:
...
-            "ubuntu pro tools not found.",
-            "tools-error",
-        )
     else:
-        return json.loads(completed_process.stdout)
+        p = Process(target=uastatus)
+        p.start()
+        p.join()
+
+        pro_info = q.get()
```
Perfect5th, line 111: I suggest we add a `timeout` to the `q.get()`, just to be on the safe side. Unsure what the timing should be but it can be pretty high.

See https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.get

landscape/client/manager/ubuntuproinfo.py
```diff
@@ -30,12 +33,45 @@ def get_data(self):
                           sort_keys=True)
 
 
+q = Queue()
```
Perfect5th, line 36: Having this be a global has me a tiny bit concerned. What if things start to pile up in it?

I think we should create a new `Queue` every time we would call `uastatus`, to be excessively safe.

Perfect5th (changes requested):

landscape/client/manager/ubuntuproinfo.py
david-mclain: fixed

st3v3nmw:
Hi @david-mclain,

Importing `from uaclient.status import status` at the top seems to break the snap:

```shell
$ sudo landscape-client.config --account-name ...
  File "/snap/landscape-client/x1/usr/bin/landscape-config", line 12, in <module>
    from landscape.client.configuration import main
  File "/snap/landscape-client/x1/lib/python3.12/site-packages/landscape/client/configuration.py", line 22, in <module>
    from landscape.client.broker.registration import Identity
  File "/snap/landscape-client/x1/lib/python3.12/site-packages/landscape/client/broker/registration.py", line 17, in <module>
    from landscape.client.manager.ubuntuproinfo import get_ubuntu_pro_info
  File "/snap/landscape-client/x1/lib/python3.12/site-packages/landscape/client/manager/ubuntuproinfo.py", line 12, in <module>
    from uaclient.status import status
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/uaclient/status.py", line 10, in <module>
    from uaclient import (
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/uaclient/livepatch.py", line 19, in <module>
    from uaclient.http import serviceclient
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/uaclient/http/serviceclient.py", line 8, in <module>
    from uaclient import config, http, system, util, version
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/uaclient/config.py", line 7, in <module>
    from uaclient import (
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/uaclient/apt.py", line 14, in <module>
    from apt.progress.base import AcquireProgress  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/snap/landscape-client/x1/usr/lib/python3/dist-packages/apt/__init__.py", line 38, in <module>
    apt_pkg.init_system()
apt_pkg.Error: E:Error reading the CPU table
```

PR #348, opened 2025-08-06
fix: imports for snap/core devices

david-mclain:


david-mclain:
@st3v3nmw @Perfect5th here's a temporary fix for the issues with imports, we plan on introducing a wrapper to the `uaclient` library this pulse or next pulse to ensure we don't run into any import issues when it comes to snaps and core devices. Sorry for running into the issue initially I had not anticipated this from the start but will ensure we don't run into this issue in the future.

st3v3nmw:
Thanks for the fix!

Perfect5th (approved):

PR #358, opened 2025-08-19
refactor: add wrapper for uaclient library calls

david-mclain:
## Manual testing:
- Run python in `landscape-client` home directory and run the following:
```
from landscape.lib.uaclient import get_pro_status
from pprint import pprint

pro_status = get_pro_status()
pprint(pro_status)
```
Verify output contains pro information

- Set in `/l/c/__init__` to have `IS_SNAP = True` and `IS_CORE = True` and verify the exception is caught and returns `{}`

### DIFF
--- a/landscape/client/manager/tests/test_ubuntuproinfo.py
+++ b/landscape/client/manager/tests/test_ubuntuproinfo.py
@@ -113,7 +113,10 @@ class UbuntuProInfoTest(LandscapeTest):
         self.assertTrue(result["attached"])
 
     def test_serializable(self):
-        """Tests calling `pro status` when it is not installed."""
+        """
+        Tests calling `pro status` when datetimes are not serialized
+        and verifies it gets serialized.
+        """
         plugin = UbuntuProInfo()
         self.manager.add(plugin)
 
@@ -145,7 +148,10 @@ class UbuntuProInfoTest(LandscapeTest):
         )
 
     def test_serializable_already(self):
-        """Tests calling `pro status` when it is not installed."""
+        """
+        Tests calling `pro status` with serialized data and verifies
+        data does not change.
+        """
         plugin = UbuntuProInfo()
         self.manager.add(plugin)
 
--- a/landscape/lib/tests/test_uaclient.py
+++ b/landscape/lib/tests/test_uaclient.py
@@ -1,6 +1,6 @@
 from unittest import TestCase, mock
 
-from landscape.lib.uaclient import get_pro_status, attach_pro, detach_pro
+from landscape.lib.uaclient import get_pro_status
 
 
 class TestUAClientWrapper(TestCase):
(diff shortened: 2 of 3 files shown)


landscape/lib/uaclient.py
```diff
@@ -0,0 +1,25 @@
+from landscape.client import IS_CORE
+from landscape.client import IS_SNAP
+
+if not IS_SNAP and not IS_CORE:
+    from uaclient.status import status
+    from uaclient.config import UAConfig
```
Perfect5th, line 6: Although this seems like it makes it obvious that the function below will not work for the snap and core, I think it's still a good idea to make sure it doesn't completely crash if for some reason it were imported and called.

Maybe have `get_pro_status` emit a warning and return an empty response?

landscape/lib/tests/test_uaclient.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_get_pro_status(self, mock_uaconfig, mock_status):
+        mock_uaconfig.return_value = None
+        mock_status.return_value = self.mock_status_value
+
+        pro_status = get_pro_status()
+
+        self.assertEqual(self.mock_status_value, pro_status)
+
+    def test_attach_pro(self):
+        self.assertTrue(True)
```
Perfect5th, line 41: Comments saying these are stub tests with `TODO` are probably more clear than this.

Perfect5th (changes requested):
A couple minor things. Works well.

Perfect5th (approved):
LGTM, thanks, just need to improve test coverage.

david-mclain:
Added some extra things for bug fix outside of just creating the wrapper. There was a bug with data serialization with the datetimes provided in the response so that is included in this

landscape/lib/tests/test_uaclient.py
```diff
@@ -0,0 +1,64 @@
...
+        with self.assertLogs() as log:
+            get_pro_status()
+
+        self.assertEqual(1, len(log.output))
+        self.assertIn(
+            "Tried to use uaclient in SNAP or CORE environment, skipping call",
+            log.output[0],
+        )
+
+    def test_attach_pro(self):
```
abehnia, line 52: Remove these two since they don't do much right now.

landscape/lib/tests/test_uaclient.py
```diff
@@ -0,0 +1,64 @@
+from unittest import TestCase, mock
+
+from landscape.lib.uaclient import get_pro_status, attach_pro, detach_pro
+
+
+class TestUAClientWrapper(TestCase):
```
abehnia, line 6: Any other tests that we can think of functional wise?

landscape/client/manager/tests/test_ubuntuproinfo.py
```diff
@@ -97,6 +112,71 @@ def test_ubuntu_pro_info(self):
         result = json.loads(messages[0]["ubuntu-pro-info"])
         self.assertTrue(result["attached"])
 
+    def test_serializable(self):
+        """Tests calling `pro status` when it is not installed."""
```
abehnia, line 116: This is a bit of nit, you don't have to do it. I prefer tests to mention what they test and what they _expect_. Here this would change to _Calling `pro status` when it is not installed should do X_. The description is probably wrong anyway.

landscape/client/manager/tests/test_ubuntuproinfo.py
```diff
@@ -97,6 +112,71 @@ def test_ubuntu_pro_info(self):
...
+            json.dumps(
+                expected,
+                separators=(",", ":"),
+                default=str
+            ),
+            messages[0]["ubuntu-pro-info"]
+        )
+
+    def test_serializable_already(self):
+        """Tests calling `pro status` when it is not installed."""
```
abehnia, line 148: The description for both this test and the previous one are the same.

landscape/lib/tests/test_uaclient.py
david-mclain: Not really, I added a new test and catch for a general exception though

abehnia (approved):
:shipit:

PR #359, opened 2025-08-27
fix: rename ssl-public-key parameter

JacobisEpic:
### Context
`ssl-public-key` is misleading. CA certificate is used to verify a Landscape server. To be consistent with other charms (e.g. OpenStack), `ssl-ca` will be used instead. `ssl-public-key` will be supported until 26.04.

### Jira
https://warthogs.atlassian.net/browse/LNDENG-2917

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+landscape-client (25.08.2-0landscape0) jammy; urgency=medium
+
+  * build: added packaging to build-depends
+
+ -- David McLain <david.mclain@canonical.com>  Thu, 14 Aug 2025 23:44:58 +0000
+
 landscape-client (25.08.1-0landscape0) jammy; urgency=medium
 
   * build: added setuptools to build-depends
--- a/debian/control
+++ b/debian/control
@@ -5,7 +5,7 @@ Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
 XSBC-Original-Maintainer: Landscape Team <landscape-team@canonical.com>
 Build-Depends: debhelper (>= 11), po-debconf, libdistro-info-perl,
                dh-python, python3-dev, python3-setuptools,
-               lsb-release, gawk, net-tools,
+               lsb-release, gawk, net-tools, python3-packaging,
                python3-apt, python3-twisted, python3-configobj,
                python3-pycurl, python3-netifaces, python3-yaml,
                ubuntu-advantage-tools, locales-all, python3-dbus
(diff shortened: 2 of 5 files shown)


landscape/client/attachments.py
```diff
@@ -37,9 +38,12 @@ async def save_attachments(
             # Backward-compatibility with inline attachments.
             data = attachment_id.encode()
         else:
```
david-mclain, line 40: Need test cases, make sure to include test cases for every piece of branching logic. Tests are located in a sub-directory of where the file is so for example in this case `landscape/client/tests/test_attachments.py`

landscape/client/attachments.py
```diff
@@ -37,9 +38,12 @@ async def save_attachments(
...
+            cainfo = getattr(config, "ssl_ca", None) or getattr(config, "ssl_public_key", None)
+            if not getattr(config, "ssl_ca", None) and getattr(config, "ssl_public_key", None):
+                logging.warning("`ssl_public_key` is deprecated; use `ssl_ca` instead.")
             data = await fetch_async(
```
david-mclain, line 44: is there a way we could avoid using `getattr` here to avoid any potential issues with if one does not exist

landscape/client/broker/ping.py
```diff
@@ -139,9 +140,12 @@ def get_interval(self):
 
     def start(self):
         """Start pinging."""
+        cainfo = getattr(self._config, "ssl_ca", None) or getattr(self._config, "ssl_public_key", None)
```
david-mclain, line 143: Same as above, tests and try to avoid `getattr` if possible

landscape/client/broker/service.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, config):
         self.transport = self.transport_factory(
             self.reactor,
             config.url,
-            config.ssl_public_key,
+            getattr(config, "ssl_ca", None) or getattr(config, "ssl_public_key", None),
```
david-mclain, line 59: Same as above

landscape/client/package/reporter.py
```diff
@@ -234,10 +234,12 @@ def fetch_error(failure):
                 proxy = self._config.get("https_proxy")
             else:
                 proxy = self._config.get("http_proxy")
-
+            cainfo = getattr(self._config, "ssl_ca", None) or getattr(self._config, "ssl_public_key", None)
```
david-mclain, line 237: Same as above

landscape-client.conf
```diff
@@ -1,11 +1,11 @@
 [client]
 bus = session
-computer_title = John's PC
-account_name = onward
-registration_key =
-url = http://localhost:8080/message-system
+computer_title = jacob-chin-canonical
```
david-mclain, line 3: Make sure not to include your personal configurations

david-mclain (changes requested):
Overall just needs test cases to ensure everything functions as expected with unit tests and generally we want to manually test these as well to ensure it will function as expected at runtime, you'll want to include those manual testing instructions in the PR description

PR #360, opened 2025-08-27
fix: rename ssl-public-key parameter

JacobisEpic:
### Context
`ssl-public-key` is misleading. CA certificate is used to verify a Landscape server. To be consistent with other charms (e.g. OpenStack), `ssl-ca` will be used instead. `ssl-public-key` will be supported until 26.04.

### Jira
https://warthogs.atlassian.net/browse/LNDENG-2917

## Server Setup
1) Run the `landscape-server` in a dev_lxc container.
2) Ensure that `enable_password_authentication = true` is placed under the `[global]` scope.
3) Run the container, `./dev/dev_lxc shell jammy`
4) Run the server './dev/racecar'

## Client Setup
1) Run `lxc list` and copy the IP address of your server container
2) Edit the `landscape-client.conf` file. 
3) Replace `computer_title`, `account_name`, `url`, and `ping_url`
4) `url = https://<Your Server Container IP>/message-system`
5) Download the cert from the server `echo | openssl s_client -connect <Your Server Container IP>:443 | openssl x509 | sudo tee /etc/landscape/server.pem`

## Manual Testing
1) Run the client with `sudo ./scripts/landscape-config --ssl-public-key =/etc/landscape/server.pem --silent` to test that `ssl-public-key` is still supported
2) Run the client with `sudo ./scripts/landscape-config --ssl-ca =/etc/landscape/server.pem --silent` to test that `ssl-ca` is also supported


david-mclain (changes requested):
Left a few comments that generally apply in most test files you changed

landscape/client/tests/test_deployment.py
```diff
@@ -224,6 +224,11 @@ def test_ssl_public_key_default(self):
         options = self.parser.parse_args([])
         self.assertEqual(options.ssl_public_key, None)
 
+    def test_ssl_ca_default(self):
```
david-mclain, line 227: We should also have a test for the option not just default, see `test_ssl_public_key_option`

david-mclain (changes requested):

david-mclain (commented):
Manual testing went well, looking into code and everything soon

david-mclain (approved):
LGTM :boat: :fr: :ship: :it: :shipit:

PR #364, opened 2025-09-02
Bug fix: add id for devmode snaps for server indexing

mcw-work:
https://bugs.launchpad.net/landscape/+bug/2121798

### DIFF
--- a/landscape/client/manager/snapmanager.py
+++ b/landscape/client/manager/snapmanager.py
@@ -292,6 +292,28 @@ class SnapManager(BaseSnapManager):
             self._send_snap_update,
         )
 
+    def _get_serial(self):
+        """
+        Returns device's serial number by parsing the serial assertion.
+
+        If the serial assertion cannot be retrieved, returns "no-serial".
+        """
+        serial = "no-serial"
+
+        try:
+            serial_bytes = snap_http.get_assertions("serial").result
+            serial_assert = serial_bytes.decode("utf-8")
+        except SnapdHttpException as e:
+            logging.error(f"Unable to get serial assertion: {e}")
+            return serial
+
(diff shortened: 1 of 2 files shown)


dilyn-corner:
The collision is a bit more sinister I think because any locally installed snap whose name matches to a snap in the store will print the store snap's info etc if you query it, even though the local snap may have nothing to do with the snap in the store...

mcw-work:
How about this as a suggestion @Perfect5th - for devmode snaps we use <device-serial>-<snap name> for the id. I have parsed the device serial from the serial assertion but perhaps this should be done and added to the device information (with brand and model details?).

landscape/client/manager/snapmanager.py
```diff
@@ -292,6 +292,27 @@ def run(self):
...
+
+        try:
+            serial_assert = snap_http.get_assertions("serial").result
+        except SnapdHttpException as e:
+            logging.error(f"Unable to get serial assertion: {e}")
+            return serial
+
+        for line in serial_assert.splitlines():
+            if line.startswith("serial:"):
+                serial = line.split("serial:", 1)[1].strip()
```
Perfect5th, line 311: In the unlikely event that `line` is just `"serial:"`, then this function will return `""` and the snap ID will be `-{snap_name}`.

Are we okay with that. How likely is it? Should we guard against it?

Perfect5th (changes requested):

landscape/client/manager/snapmanager.py
Perfect5th: A quick fix is to end with:
```python
return serial or "no-serial"
```

Perfect5th (approved):

PR #371, opened 2025-09-17
feat: add weekly integration testing for uaclient/pro

Perfect5th:
This change introduces an integration test for `uaclient`, which drives our Ubuntu Pro management features.

It introduces pytest, mainly to take advantage of its "mark" capabilities, which keeps integration test runs separate from unit test runs without needing to introduce modularization or pattern-matching on test identifiers.

To run the integration test, install pytest and run it using the integration marker:

```
sudo apt install python3-pytest
pytest -m integration
```

Be aware that on jammy, the pytest executable installed by python3-pytest is actually `py.test-3` for some reason.

Tests will fail unless you've set the environment variable `TEST_PRO_TOKEN` to a valid, "free" Ubuntu Pro token, which you can get from [ubuntu.com/pro](https://ubuntu.com/pro). They will also fail unless your user has pro management permissions, which usually means being root.

A successful run usually looks like:

```
sudo TEST_PRO_TOKEN="MyToken" pytest -m integration
=============================================================== test session starts ================================================================
platform linux -- Python 3.10.12, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
rootdir: /home/ubuntu/landscape-client, configfile: pyproject.toml
collected 2668 items / 2667 deselected / 1 selected                                                                                                

landscape/lib/tests/test_uaclient_integration.py .                                                                                           [100%]

======================================================= 1 passed, 2667 deselected in 12.43s ========================================================
```

More integration tests can be added in the future for other integrations.


Perfect5th:
Heads up, there's a handful of class renamings: pytest automagically collects classes in files that start with `test_` that are named `Test.*`.

I replaced all instances with `Mock.*`.

PR #401, opened 2026-01-06
refactor: inline broadcast_event in broker server

srunde3:
## Context

The `@event` decorator changed the function signature of the methods it decorated, and this obfuscated the arguments that the event should accept. IMO it is an unnecessary abstraction.

This is a small step towards better understanding the event system in the client. In the future I'd like to type the events (e.g., enum) in the future so that we can better reason about event senders and receivers.


landscape/client/broker/server.py
```diff
@@ -315,13 +307,13 @@ def schedule_reactor_stop(ignored):
 
         return clients_stopped.addBoth(schedule_reactor_stop)
 
-    @event
-    def resynchronize(self):
+    def resynchronize(self, scopes=None):
```
jmucc, line 310: Why did we add scopes? Is it just for the one test? It didn't say "scope" or anything in it.

jmucc (commented):
One question

landscape/client/broker/server.py
srunde3: This was part of the reason for my change - the `@event` decorator obscured the arguments that these methods could actually take. The handler for this message is in `BrokerClientPlugin`:

```python
class BrokerClientPlugin:

    ...
   
    def register(self, client: "BrokerClient"):
        self.client = client
        self.client.reactor.call_on("resynchronize", self._resynchronize)  # <-- here
        deferred = self.client.broker.get_session_id(scope=self.scope)
        deferred.addCallback(self._got_session_id)

    ...

    def _resynchronize(self, scopes=None):
        """
        Handle the 'resynchronize' event.  Subclasses should do any clear-down
        operations specific to their state within an implementation of the
        L{_reset} method.
        """
        ...
```

`scopes` was always handled by `BrokerClientPlugin`s, it just wasn't obvious that the BrokerServer could emit it.

jmucc (approved):
LGTM

PR #419, opened 2026-03-19
fix: restore tests for amd64v3

jmucc:
This is blocking the 26.04 archive release for landscape-client because these tests fail on amd64v3, which means the package cannot build. The tests that were failing were for edge cases that should never occur in practice, and they were causing segmentation faults with amd64v3's optimized instructions. The tests have been modified to explicitly mock those edge cases.

I am triggering a launchpad build on my test PPA to see if the builds for amd64v3 go through properly, I will open the PR for review once that succeeds. 

https://launchpad.net/~joey-mucci/+archive/ubuntu/resolute-test-2/+sourcepub/18177769/+listing-archive-extra


wck0 (approved):
LGTM

PR #423, opened 2026-03-25
build: release 26.02.3 for jammy

jmucc:
Creating another patch release so the PPAs can have the patch applied in the archive release.


Perfect5th (approved):
LGTM

PR #424, opened 2026-04-07
ci: add ppa deploy

jmucc:
We should automate beta releases for landscape-client to `self-hosted-beta` as per internal documentation

Successful run for refactored build jammy: https://github.com/canonical/landscape-client/actions/runs/24107083785

We can test the ppa-deploy on the first beta-release. The actual ppa-upload logic is copy-paste from server. As far as I am aware there is no other validation on the landscape-client deb artifact except for building, so that's the only prereq for a beta upload for now.


wck0 (approved):
LGTM

PR #428, opened 2026-04-10
build: release 26.08~beta.1.1 for jammy

jmucc:
We need another beta release because the last one was broken and consumers of our self-hosted-beta PPA are experiencing interruptions in their CI. Also, if it makes sense for this to be `beta.1-0landscape1` instead let me know.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-landscape-client (26.08~beta.1.1-0landscape0) jammy; urgency=medium
+landscape-client (26.08~beta.2-0landscape0) jammy; urgency=medium
 
   * fix: snap builds & installation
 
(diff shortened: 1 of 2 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+landscape-client (26.08~beta.1.1-0landscape0) jammy; urgency=medium
```
Perfect5th: I think it's fine for this to be `beta.2`. Any reason not to?

jmucc: No not really. I was thinking to keep the releases tied to a pulse but that introduces more complexity than its worth.

Perfect5th (approved):

PR #429, opened 2026-04-14
fix: disconnect connectors on error paths to prevent socket leak (LP: #2147687)

rates37:
## Description:

Connector `disconnect()` was only called on the success path via `addCallback`, meaning any failure mid-chain left the socket connector open. Over long periods of operation, this caused a slow but steady FD leak, eventually exhausting the process's file descriptor limit with `Errno 24: Too many open files`.

## Changes:
* `usermanager.py`: Changed `addCallback` to `addBoth` for `disconnect`, ensuring cleanup on both success and error paths.
* `usermonitor.py`: Same fix as `usermanager.py`
* `watchdog.py`: Same fix as `usermanager.py`
* `server.py`: Disconnect the existing connector if a client re-registers using the same name

## Launchpad Bug:

https://bugs.launchpad.net/landscape-client/+bug/2147687

### DIFF
--- a/landscape/client/monitor/tests/test_usermonitor.py
+++ b/landscape/client/monitor/tests/test_usermonitor.py
@@ -5,7 +5,7 @@ from twisted.internet.defer import fail
 
 import landscape.client.monitor.usermonitor
 from landscape.client.amp import ComponentPublisher
-from landscape.client.manager.usermanager import UserManager
+from landscape.client.manager.usermanager import RemoteUserManagerConnector, UserManager
 from landscape.client.monitor.usermonitor import RemoteUserMonitorConnector, UserMonitor
 from landscape.client.tests.helpers import LandscapeTest, MonitorHelper
 from landscape.client.user.tests.helpers import FakeUserProvider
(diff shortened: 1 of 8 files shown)


landscape/client/monitor/tests/test_usermonitor.py
```diff
@@ -534,6 +534,45 @@ def got_result(result):
...
+        result = connector.connect()
+        result.addCallback(lambda remote: remote.detect_changes())
+
+        def assert_disconnect(ignored):
+            self.assertTrue(CallTracker.called)
+
+        # The detect_changes call returns the RPC deferred which swallows the
+        # errback internally into a SUCCESS of None due to
+        # `result.addErrback(lambda f: self._detect_changes([], operation_id))`.
+        result.addBoth(lambda x: connector.disconnect())
```
jmucc, line 572: nit: Why is this needed? For me the test passes with these 4 lines removed.

jmucc (commented):
The changes looks really good. I just need a little bit of time to verify the fix with the scripts in provided in the LP bug.

jmucc (approved):
I was able to verify the fix for the first two scripts provided in the LP bug, and the changes look minimal. LGTM.

PR #433, opened 2026-05-15
client: set User-Agent for ping requests

andreweskander-2:
## Summary

Set an explicit `User-Agent` header for ping requests.

`exchange.py` already identifies itself using:

```python id="v4af8t"
User-Agent: landscape-client/{VERSION}
```

but `ping.py` currently falls back to the default PycURL/libcurl generated user-agent.

This results in inconsistent client identification across Landscape endpoints and may cause `/ping` requests to be filtered differently by reverse proxies or WAFs.

Use the same Landscape-specific `User-Agent` for ping requests to align behavior with `/message-system`.

## Testing

```bash id="bny9wq"
pytest landscape/client/broker/tests/test_ping.py
```

### DIFF
--- a/landscape/client/broker/ping.py
+++ b/landscape/client/broker/ping.py
@@ -43,6 +43,7 @@ except ImportError:
 from twisted.internet import defer
 from twisted.python.failure import Failure
 
+from landscape import VERSION
 from landscape.lib import bpickle
 from landscape.lib.fetch import fetch
 from landscape.lib.log import log_failure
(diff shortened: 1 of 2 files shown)


andreweskander-2:
relevant troubleshooting and more details can be found [here](https://discourse.canonical.com/t/landscape-deployment-cloudflare-waf-blocking-landscape-heartbeats-error-1010/7457).

Perfect5th (approved):
LGTM. Thanks for this.

PR #435, opened 2026-06-15
feat(snap): add core26 support

st3v3nmw:
This PR introduces `core26` support to the landscape-client snap. 

The `core26` base [no longer bundles a Python interpreter](https://forum.snapcraft.io/t/core26-bundled-python-changes/50178) hence the refactors in the `snapcraft.yaml` to [bundle a Python runtime with the snap](https://documentation.ubuntu.com/snapcraft/9.0/how-to/change-bases/change-from-core24-to-core26/#repackage-python) and use that instead of the interpreter previously bundled with the base. The standard `plugin: python` virtualenv-based build is also broken in core26, complaining that `ensurepip` is not available despite it being present, so this PR overrides build with a direct `pip install --target` similar to what we used to have last year: https://github.com/canonical/landscape-client/blob/30995b3c4abace0f89d17eaeba7ac8346ad4e43b/snap/snapcraft.yaml#L111-L113

The `#!/usr/bin/python3` shebangs in the primed scripts are also rewritten to `#!/usr/bin/env python3` so that script execution resolves to the snap's own bundled interpreter rather than the absent base one.

---

The snap is [building successfully](https://github.com/canonical/landscape-client/actions/runs/27683193051) across all 3 core bases we now support. And I've tested it against staging:

image

<br/>

Lastly, the artefact size has increased a little, 15M on previous core24 builds to 18M now.. 19M on core26. The build isn't happy with `python3-minimal` and friends..

### DIFF
--- a/snap/snapcraft.yaml.j2
+++ b/snap/snapcraft.yaml.j2
@@ -1,9 +1,11 @@
 {# this is a template, generate the actual snapcraft.yaml with `make snap-yaml` #}
-{%- set python_version = "3.12" if base == "core24" else "3.10" -%}
+{%- set python_versions = {"core22": "3.10", "core24": "3.12", "core26": "3.14"} -%}
+{%- set python_version = python_versions[base] -%}
+{%- set supported_archs = ["amd64", "arm64", "armhf", "ppc64el", "s390x", "riscv64"] -%}
 
 name: landscape-client
 base: {{ base }}
-version: '26.02'
+version: '26.08'
 icon: snap/gui/landscape-logo-256.png
 website: https://ubuntu.com/landscape
 license: GPL-2.0-only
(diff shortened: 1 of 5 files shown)


snap/snapcraft.yaml.j2
```diff
@@ -133,23 +137,27 @@ parts:
...
-    override-stage: |
+    override-prime: |
       craftctl default
 
       # Copy the scripts over
       mkdir -p ${SNAPCRAFT_PRIME}/usr/bin/
       cp -r ${SNAPCRAFT_PART_SRC}/scripts/. ${SNAPCRAFT_PRIME}/usr/bin/
+
+      # Rewrite shebangs to use the snap-local python3
+      sed -i '1s|#!/usr/bin/python3|#!/usr/bin/env python3|' ${SNAPCRAFT_PRIME}/usr/bin/landscape-*
```
wck0, line 159: Is there any reason we wouldn't want the shebangs to be `/usr/bin/env python3` everywhere?

wck0 (approved):
Approving, but one question below.

snap/snapcraft.yaml.j2
st3v3nmw: I suppose we can do that.. I just didn't want to touch the deb 😅.

st3v3nmw:
Hey @wck0, @Perfect5th, can I get a merge for this please?

PR #437, opened 2026-07-16
fix: snap release workflow

st3v3nmw:
The [org-level default workflow privileges were updated](https://discourse.canonical.com/t/upcoming-change-to-github-actions-settings/7286) from "read and write" to "read". This PR fixes the "Update release branches" workflow with the right permissions to write to the release branches.

### DIFF
--- a/.github/workflows/snap-builds.yml
+++ b/.github/workflows/snap-builds.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
           submodules: true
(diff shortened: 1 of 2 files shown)


.github/workflows/snap-builds.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
```
wck0: Github actions are generally trusted, but I think it's better to protect ourselves against tag hijacking wherever possible.

Can you change this line, and the similar one below, to reference the SHA instead, and leave a comment that if's v7?

https://github.com/actions/checkout/releases/tag/v7.0.0

wck0 (changes requested):
One request to be a bit more paranoid.

.github/workflows/snap-builds.yml
st3v3nmw: Done.

wck0 (approved):
LGTM

Perfect5th (approved):

PR #447, opened 2026-07-22
refactor: cap error count for exponential backoff

wck0:
Previously, the error count would grow without bound, even though the max delay effectively capped the delay. Message exchange has a hard coded max delay of 7200 seconds (2 hours). That means if the client's message exchange failed 10 times in a row, it could be more than a day before message exchange goes back to normal.

With these changes, the error count is capped to a level based on the configured start and max delay values.

### DIFF
--- a/landscape/lib/backoff.py
+++ b/landscape/lib/backoff.py
@@ -43,8 +43,7 @@ class ExponentialBackoff:
         Increases the error count, capped at the limit needed
         to reach the maximum delay.
         """
-        if self._error_count < self._max_effective_error_count:
-            self._error_count += 1
+        self._error_count = min(self._error_count + 1, self._max_effective_error_count)
 
     def get_delay(self) -> int:
         """
--- a/landscape/lib/backoff.py
+++ b/landscape/lib/backoff.py
@@ -1,3 +1,17 @@
+"""
+Exponential Backoff Utility.
+
+This module provides an implementation of an exponential backoff algorithm.
+
+It features:
+- Exponentially increasing delays based on consecutive error counts.
+- A configurable maximum delay ceiling to prevent unbounded waits.
+- Gradual recovery by stepping down the error count on success.
+- Randomized jitter to prevent "thundering herd" scenarios when multiple
+  clients are retrying simultaneously.
+"""
+
+import math
 import random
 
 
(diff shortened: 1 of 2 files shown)


landscape/lib/backoff.py
```diff
@@ -36,12 +57,12 @@ def get_delay(self):
                 5      60 (max)
         """
         if self._error_count:
-            delay = (2 ** (self._error_count - 1)) * self._start_delay
+            delay = self._start_delay << (self._error_count - 1)
```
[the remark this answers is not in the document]
wck0: Yup, sure is. Intentional, too.

landscape/lib/tests/test_backoff.py
```diff
@@ -34,3 +36,52 @@ def test_decreased_when_maxed(self):
...
+
+    def test_error_count_cap(self):
+        """Test that the internal error count never exceeds the effective max"""
+        backoff_counter = ExponentialBackoff(5, 60)
+
+        # Increase well beyond the effective max
+        for _ in range(15):
+            backoff_counter.increase()
+
+        self.assertEqual(backoff_counter._error_count, 5)
```
jmucc: Do we care that  the `_error_count` is equal to 5? I think we only care that `_error_count == _max_effective_error_count` and that `_max_effective_error_count` is appropriate. We could check if `get_delay` only reaches `max_delay` for `_max_effective_error_count` and not `_max_effective_error_count - 1`.

landscape/lib/backoff.py
```diff
@@ -7,23 +21,32 @@ class ExponentialBackoff:
...
-    def increase(self):
-        """Increases error count but not higher than gives the max delay"""
-        if self.get_delay() < self._max_delay:
+    def increase(self) -> None:
+        """
+        Increases the error count, capped at the limit needed
+        to reach the maximum delay.
+        """
+        if self._error_count < self._max_effective_error_count:
             self._error_count += 1
```
jmucc: nit: Is there a reason these two helpers use different formats?

jmucc (commented):
Let me know what you think about the test thing.

landscape/lib/tests/test_backoff.py
wck0: We kinda do. We don't want the error count to grow without bound. Capping the delay is also desirable but orthogonal.

landscape/lib/backoff.py
wck0: Nope. I'll make them match.

landscape/lib/tests/test_backoff.py
wck0: Oh, I see what you mean. Yeah, I'll remove the extra check.

jmucc (approved):
LGTM

PR #448, opened 2026-07-24
fix: broker stale socket recovery

bugraaydogar:
When the watchdog SIGKILLs a broker, its unix socket (broker.sock) is left behind with no live listener. The next broker fails to bind it (CannotListenError: Address already in use) and crashes during startup — before logging, so it only shows in the journal. After 5 quick crashes the watchdog gives up and exits 0, so systemd never restarts the client. Device stays online but goes dark on Landscape until snap restart.

### DIFF
--- a/landscape/client/amp.py
+++ b/landscape/client/amp.py
@@ -11,8 +11,12 @@ let the various services connect to each other in an easy and idiomatic way,
 and have them respond to standard requests like "ping" or "exit".
 """
 
+import errno
 import logging
 import os
+import socket as socket_module
+
+from twisted.internet import error as tw_error
 
 from landscape.lib.amp import (
     MethodCallClientFactory,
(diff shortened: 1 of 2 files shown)


st3v3nmw (commented):
I've verified that this fix works per the bug reproduction in https://warthogs.atlassian.net/browse/LNDENG-4873?focusedCommentId=1164580.

I don't have merge rights on this repository, but I'll build and upload the fix to the `24/edge` channel for the `amd64`, `arm64`, & `armhf` architectures. This will hold the fort until the landscape team is back.

https://launchpad.net/~st3v3nmw/st3v3nmw-craft-remote-build/+snap/snapcraft-landscape-client-84973f5619ec46b8e19a008cdad6e1d8

st3v3nmw:
Future reviewers, please see the discussion in [this channel](https://chat.canonical.com/canonical/channels/sf-00444718).

prash813:
@bugraaydogar  @st3v3nmw , any specific reason why the lock files related to respective sockets has not been deleted.
I can also see the error in one of the log files
```
raise error.CannotListenError(None, self.port, "Cannot acquire lock")
```
which I think is related to presence of lock file.

david-mclain (commented):
Hello, I reviewed this and everything is good was just missing some codecov. I added the test cases required and will have @Perfect5th ensure everything is good.

Perfect5th (approved):
LGTM, thanks!

PR #452, opened 2026-08-12
fix: client config file permissions in snap (LP# 2161005)

wck0:
## Manual testing instructions

1. Install the snap from the store and register with Landscape Server.

    ```bash
    sudo snap install landscape-client
    ```

1. Check the file permissions are 755:

    ```bash
    ls -la /var/snap/landscape-client/common/etc/landscape-client.conf
    ```

1. Register with Landscape Server

    ```bash
    sudo landscape-client.config
    ```

1. Pack the snap from this branch.

    ```bash
    make snap
    ```

1. Install it.

    ```bash
    sudo snap install --dangerous ./landscape-client_26.08_amd64.snap
    ```

1. Check the file permissions are 600.

    ```bash
    ls -la /var/snap/landscape-client/common/etc/landscape-client.conf
    ```

1. Kick off an activity for the client and make sure it works.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -77,6 +77,11 @@ for snapctl_key, landscape_key, mapping_fn in [*legacy_config_entries, *config_e
 
 config.write()
 
+# May contain sensitive values (e.g. registration_key); keep it locked down.
+# config.write() doesn't reset permissions on an existing file, but this is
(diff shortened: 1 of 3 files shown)


jmucc (approved):
LGTM

